PR #24, opened 2025-10-30
Develop

ruedigerp:


.github/workflows/go-test.yaml
```diff
@@ -42,26 +46,26 @@ jobs:
...
+  #   - name: lint and test
+  #     run: |
+  #       cd src
+  #       go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
 
-        go mod download
+  #       go mod download
 
-        golangci-lint run --timeout 10m
+  #       golangci-lint run --timeout 10m
```
lukashankeln: i dont think we should disable the linter here

(we can ignore specific cases if needed, but disabling it at all seems wrong)

src/internal/renovate/discoveryAgent.go
```diff
@@ -57,21 +68,11 @@ func (e *discoveryAgent) discoverIntern(ctx context.Context, job *api.RenovateJo
...
-func (e discoveryAgent) WaitForDiscoveryJob(ctx context.Context, job *api.RenovateJob) ([]string, error) {
-	// lock based on the renovatejob
-	name := job.Fullname()
-	lock := e.syncer[name]
-	if lock == nil {
-		lock = &sync.RWMutex{}
-		e.syncer[name] = lock
-	}
-	lock.RLock()
-	defer lock.RUnlock()
```
lukashankeln, line 69: the read lock needs to be here, otherwise we can get into issues, if the job gets recreated at the same time

.github/workflows/go-test.yaml
ruedigerp: Fixed some Tests and Code and golangcli lint reenabled. 
https://github.com/mogenius/renovate-operator/actions/runs/18947130441

src/internal/renovate/discoveryAgent.go
ruedigerp: That is now handled by getDiscoveryJobStatusInternal. The lock is also set there.
Is that okay with you?

Line: 122-...
`func (e *discoveryAgent) getDiscoveryJobStatusInternal(ctx context.Context, job *api.RenovateJob) (api.RenovateProjectStatus, error) {
	// lock based on the renovatejob
	name := job.Fullname()
	lock := e.syncer[name]
	if lock == nil {
		lock = &sync.RWMutex{}
		e.syncer[name] = lock
	}
	lock.RLock()
	defer lock.RUnlock()`

PR #70, opened 2025-12-16
feat: add (HTTP)route support to helm chart

RonaldPhilipsen:
With ingress-nginx being deprecated soon it'd be nice to be able to template out httproutes.

### DIFF
--- a/charts/renovate-operator/templates/route.yaml
+++ b/charts/renovate-operator/templates/route.yaml
@@ -1,6 +1,24 @@
 {{- with .Values.route }}
 {{- range $routeName, $routeConfig := . }}
 {{- if (default true $routeConfig.enabled) }}
+{{- $rules := $routeConfig.rules }}
+{{- if not $rules }}
+{{- $backendRefs := list (dict
+    "name" (include "renovate-operator.fullname" $)
+    "port" 8081
+    "weight" 1
+) }}
+{{- if $.Values.webhook.enabled }}
+{{- $backendRefs = append $backendRefs (dict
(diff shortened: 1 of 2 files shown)


charts/renovate-operator/templates/route.yaml
```diff
@@ -0,0 +1,60 @@
...
+      matches:
+        {{- toYaml $rule.matches | nindent 8 }}
+      {{- end }}
+      {{- if $rule.backendRefs }}
+      backendRefs:
+        {{- toYaml $rule.backendRefs | nindent 8 }}
+      {{- else }}
+      backendRefs:
+        - name: {{ include "renovate-operator.fullname" $ }}
+          port: 8081
```
lukashankeln: We also have a second ingress for webhooks. (see webhook key in the values)

That one is hosted with a different port, so hard coding the ui port here would make it difficult to create two routes for both ingresses.

Could we change this so that booth ingresses are supported?

Maybe adding them in the values or something

RonaldPhilipsen: yeah, that makes sense to me.

RonaldPhilipsen: Hmm, thinking back at this you're already able to configure the webhook with the same set of rules, you just 

```yaml

    rules:
      - name: "webui"
        backendRefs:
          - name: {{ include "renovate-operator.fullname" $ }}
            port: 8081
            weight: 1
      - name: "webui"
        backendRefs:
          - name: {{ include "renovate-operator.fullname" $ }}
            port: 8082
            weight: 1
```

Issue #35, opened 2025-11-12
[Metrics] Adding metrics that describe your renovate runs


lukashankeln:
We should add some more metrics to observe issues in our renovate runs. e.g.

- Project Count per RenovateJob
- Successful runs per RenovateJob & Proejct
- Failed runs per RenovateJob & Proejct
  - Could be used for custom Alert via Prometheus rule

pascal-hofmann:
Hi,

maybe https://github.com/raffis/renovate-metrics could be integrated? It reads renovates debug logs in json format, derives metrics and sends them to a pushgateway. It already has a nice dashboard.

Would you accept a PR for a renovate-metrics integration?

Cheers
Pascal

lukashankeln:
> Hi,
> 
> maybe https://github.com/raffis/renovate-metrics could be integrated? It reads renovates debug logs in json format, derives metrics and sends them to a pushgateway. It already has a nice dashboard.
> 
> Would you accept a PR for a renovate-metrics integration?
> 
> Cheers Pascal

That Project seems to have some nice metrics. In which way would you propose to integrate this?

pascal-hofmann:
My first naive thought was to just call out to it (like we currently do after our renovate runs triggered by k8s cronjobs - we're not using this operator yet) and let it handle the metrics forwarding. But maybe it makes more sense to properly integrate it (if possible) or worst case take the relevant parts?

The maintainer usually responds really quickly, maybe we could work with him to get renovate-metrics work both standalone and as a pluggable library?

lukashankeln:
We added a lot of new metrics with #449 
Closing this Issue for now. If any other metrics are missing, we should created dedicated issues for those.

PR #92, opened 2026-02-03
feat: add metrics for Renovate job failures and dependency issues

tibuntu:
- Introduced new Prometheus metrics: `renovate_operator_run_failed` and `renovate_operator_dependency_issues` to track job failures and dependency issues.
- Updated Grafana dashboard configuration to visualize these metrics.
- Enhanced Renovate log parsing to detect dependency issues and update metrics accordingly.
- Added tests to ensure correct log format handling and metric updates.

Kind of related to: #35

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -11,6 +11,44 @@ metrics:
 
 ## Exported Metrics
 
-| Name                                       | Type    | Description                                | Labels                                                    |
-|--------------------------------------------|---------|--------------------------------------------|-----------------------------------------------------------|
-| renovate_operator_project_executions_total | Counter | Total number of executed Renovate projects | `renovate-namespace`, `renovate-job`, `project`, `status` |
+| Name                                       | Type    | Description                                                              | Labels                                                    |
+|--------------------------------------------|---------|--------------------------------------------------------------------------|-----------------------------------------------------------|
+| renovate_operator_project_executions_total | Counter | Total number of executed Renovate projects                               | `renovate-namespace`, `renovate-job`, `project`, `status` |
+| renovate_operator_run_failed               | Gauge   | Whether the last Renovate run for this project failed (1=failed, 0=success) | `renovate-namespace`, `renovate-job`, `project`           |
+| renovate_operator_dependency_issues        | Gauge   | Whether the last Renovate run had WARN/ERROR log entries (1=issues, 0=clean) | `renovate-namespace`, `renovate-job`, `project`           |
+
+## Dependency Issues Detection
+
+The `renovate_operator_dependency_issues` metric detects issues by parsing Renovate's JSON log output. This includes:
+
+- Configuration validation warnings
+- Dependency lookup failures
+- Rate limiting issues
+- Invalid `renovate.json` configuration
+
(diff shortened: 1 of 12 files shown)


docs/metrics.md
```diff
@@ -11,6 +11,44 @@ metrics:
...
+
+      - alert: RenovateDependencyIssues
+        expr: renovate_operator_dependency_issues == 1
+        for: 5m
+        labels:
+          severity: warning
+        annotations:
+          summary: "Renovate detected dependency issues for {{ $labels.project }}"
+          description: "The last Renovate run for project {{ $labels.project }} in job {{ $labels.renovate_job }} had warnings or errors. Check the Dependency Dashboard or Renovate logs for details."
+```
```
lukashankeln, line 54: Should we add these Prometheus Rules as default to the Helm-Chart?

src/internal/renovate/executor.go
```diff
@@ -196,7 +197,25 @@ func (e *renovateExecutor) reconcileProjects(ctx context.Context, renovateJob *a
...
+				// Parse logs before potential job deletion to capture dependency issues
+				hasIssues := false
+				if job != nil {
+					cp := clientProvider.StaticClientProvider()
+					if clientset, err := cp.K8sClientSet(); err == nil {
+						if logs, err := crdManager.GetLastJobLog(ctx, clientset, job); err == nil {
+							parseResult := ParseRenovateLogs(logs)
+							hasIssues = parseResult.HasIssues
+						} else {
+							e.logger.V(1).Info("failed to get logs for metrics parsing", "project", project.Name, "error", err)
```
lukashankeln: I guess this should be `e.logger.Error()`

src/internal/renovate/jobDefinitions_test.go
```diff
@@ -38,6 +38,78 @@ var (
 	}
 )
 
+func TestLogFormatJsonDefault(t *testing.T) {
```
lukashankeln: This test case is theoretically redundant. You could add the `expectEnvVar(t, djContainer, "LOG_FORMAT", "json")` to the two test cases above.

The other test case on the other hand is useful

src/internal/parser/logParser.go
lukashankeln, line 1: Can you move this to a new module? e.g. `parser` in `internal/parser` or something like that

src/metricStore/metrics.go
```diff
@@ -26,3 +42,27 @@ func CaptureRenovateProjectExecution(namespace, job, project, status string) {
...
+	if hasIssues {
+		value = 1.0
+	}
+	dependencyIssues.WithLabelValues(namespace, job, project).Set(value)
+}
+
+// DeleteProjectMetrics removes all metrics for a project that was removed from discovery
+func DeleteProjectMetrics(namespace, job, project string) {
+	runFailed.DeleteLabelValues(namespace, job, project)
+	dependencyIssues.DeleteLabelValues(namespace, job, project)
```
lukashankeln, line 67: Is it possible to delete the metric for Counters as well? In that case should we add `projectRuns` here as well?

tibuntu: Done. I've also added some tests for the metrics.

docs/metrics.md
tibuntu: I am not sure. As there is no current templating for Prometheus Rules I'd say this is out of scope.

PR #97, opened 2026-02-09
fix: using updated infos to create new jobs.

beneiltis:
When the scheduled function executed, it used the stale renovateJob — including the old container image. Repo jobs didn't have this issue because the executor re-fetches from the API server on every run.


src/controllers/renovatejob_controller.go
lukashankeln, line 32: The job is initially reloaded here on every reconciliation. I thought that would be enough, but i think it cant harm to reload the job before the discovery

PR #102, opened 2026-02-10
fix: show version in ui

beneiltis:
pretty much what the title says :-)


src/cmd/main.go
```diff
@@ -27,6 +27,8 @@ import (
 	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
+var Version = "dev"
```
lukashankeln: Should we add a short comment where this is "filled"?

beneiltis: good idea, incoming.

PR #104, opened 2026-02-10
feat: leader-election-feature

beneiltis:


src/cmd/main.go
```diff
@@ -113,6 +113,20 @@ func main() {
...
 			Optional: true,
 		},
+		{
+			Key:      "POD_NAMESPACE",
+			Optional: true,
+		},
+		{
+			Key:      "LEADER_ELECTION_ENABLED",
+			Optional: true,
+			Default:  "true",
```
lukashankeln: should the default be false here? otherwise its always on

PR #136, opened 2026-02-18
fix: crd updates will now be installed using helm upgrade

beneiltis:
addresses issue-134

hasRenovateConfig was added to the CRD schema in v2.4.0, but Helm never updates CRDs placed in the crds/ directory during helm upgrade — they are only applied on the very first install. Anyone upgrading from a pre-2.4.0 version would keep the old schema and the API server would prune + warn about the unknown field on every status update.

Fix — two-part change to [templates/renovate-operator.mogenius.com_renovatejobs.yaml](vscode-webview://12d98iame3n09difmq8e78qhs8occaukv8s2d6s66b0k9ncj1sar/charts/renovate-operator/templates/renovate-operator.mogenius.com_renovatejobs.yaml):

Moved the CRD from crds/ → templates/. Helm now applies it on both helm install and helm upgrade, keeping the schema in sync with the chart version.

Added helm.sh/resource-policy: keep annotation. This prevents helm uninstall from deleting the CRD (and all RenovateJob custom resources with it) — the standard safety net for operator CRDs.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -172,9 +172,16 @@ jobs:
             go install sigs.k8s.io/controller-tools/cmd/controller-gen@latest
 
             echo "Generating CRDs"
-            controller-gen crd paths=./src/... output:crd:dir=charts/renovate-operator/crds
+            controller-gen crd paths=./src/... output:crd:dir=charts/renovate-operator/crd
 
-            git add charts/renovate-operator/crds
+            echo "Adding Helm resource-policy annotation to CRDs"
+            for f in charts/renovate-operator/crd/*.mogenius.com_*.yaml; do
+              if ! grep -q "helm.sh/resource-policy" "$f"; then
+                sed -i 's/  annotations:/  annotations:\n    helm.sh\/resource-policy: keep/' "$f"
+              fi
+            done
+
+            git add charts/renovate-operator/crd/*.mogenius.com_*.yaml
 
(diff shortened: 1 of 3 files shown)


charts/renovate-operator/crd/renovate-operator.mogenius.com_renovatejobs.yaml
```diff
@@ -4,6 +4,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.20.1
```
lukashankeln, line 6: Will this be keped if the Generator command Runs again?

beneiltis: excelent point. let me adjust it :-)

.github/workflows/release.yaml
```diff
@@ -172,9 +172,16 @@ jobs:
             go install sigs.k8s.io/controller-tools/cmd/controller-gen@latest
 
             echo "Generating CRDs"
-            controller-gen crd paths=./src/... output:crd:dir=charts/renovate-operator/crds
+            controller-gen crd paths=./src/... output:crd:dir=charts/renovate-operator/templates
```
lukashankeln: Will a Helm Upgrade be able to handle this? Or will this get our lovely "This resource already exists and cannot be imported" error message?

PR #139, opened 2026-02-20
Add optional OIDC/OAuth2 authentication for the Web UI

ruedigerp:
### Summary

This PR adds optional OpenID Connect (OIDC) authentication to the renovate-operator Web UI. When OIDC is configured via environment variables or Helm values, all UI and API access requires login through an external identity provider. When OIDC is not configured, the operator behaves exactly as before with unrestricted access.

### Motivation

The renovate-operator Web UI allows triggering renovate runs, starting discovery jobs, and viewing logs. In environments where the UI is exposed via Ingress, it should be possible to restrict access to authenticated users only, using an existing identity provider (Keycloak, Authentik, Zitadel, Google, Azure AD, etc.).

---

#### OIDC Authentication Provider (`src/ui/oidc.go` — new file)

- **Full OIDC Authorization Code Flow** using `github.com/coreos/go-oidc/v3` and `golang.org/x/oauth2`
- **OIDC Discovery**: Automatically fetches provider configuration from `/.well-known/openid-configuration`
- **ID Token verification**: All ID tokens are cryptographically verified against the provider's JWKS
- **Stateless encrypted sessions**: Session data (email, name, expiry) is stored in AES-256-GCM encrypted cookies — no server-side session state required, works across multiple replicas without shared storage
- **Configurable session secret**: When `OIDC_SESSION_SECRET` is set, the encryption key is derived from it (via SHA-256), ensuring sessions survive pod restarts. When not set, a random key is generated at startup
- **24-hour session duration** with automatic expiry validation on every request
- **Custom HTTP client with TLS configuration**: A dedicated `http.Client` is used for all OIDC operations (discovery, token exchange, token verification) with configurable TLS settings
- **Optional TLS skip verification** (`OIDC_INSECURE_SKIP_VERIFY`) for development/testing environments with self-signed certificates
- **CSRF protection**: Login flow uses cryptographically random state parameters stored in a short-lived cookie to prevent cross-site request forgery

#### Authentication Middleware

- Wraps the entire HTTP handler when OIDC is enabled
- **`/health`** — always bypassed (Kubernetes liveness/readiness probes)
- **`/auth/*`** — always bypassed (login, callback, logout endpoints)
- **`/api/*`** — returns `401 Unauthorized` JSON response when session is missing or expired
- **All other paths (UI)** — redirects to `/auth/login` when session is missing or expired
- When OIDC is not configured, the middleware is not applied at all — zero overhead

#### New HTTP Endpoints

| Endpoint | Method | Description |
|---|---|---|
| `GET /auth/login` | GET | Initiates the OIDC login flow, sets state cookie, redirects to the identity provider |
| `GET /auth/callback` | GET | Handles the OIDC callback, validates state, exchanges authorization code for tokens, verifies ID token, creates encrypted session cookie, redirects to `/` |
| `GET /auth/logout` | GET/POST | Clears the local session cookie and redirects to the OIDC provider's `end_session_endpoint` (if configured) with `client_id` and `post_logout_redirect_uri` parameters to terminate the provider-side session. Falls back to redirecting to `/` if no logout endpoint is available |
| `GET /api/v1/auth/status` | GET | Returns the current authentication status as JSON. Always accessible (even without session). Response: `{"enabled": false}` when OIDC is not configured, or `{"enabled": true, "authenticated": true/false, "email": "...", "name": "..."}` when OIDC is active |

#### Provider Logout Support

- The `end_session_endpoint` is automatically discovered from the OIDC provider's `.well-known/openid-configuration` if available
- Can be explicitly overridden via `OIDC_LOGOUT_URL` for providers that don't advertise it in their discovery document
- Logout flow: clear local cookie → redirect to provider logout → provider redirects back to the application

---

#### `src/ui/oidc.go` (new)
Complete OIDC implementation: provider initialization, authentication middleware, login/callback/logout handlers, encrypted cookie session management (AES-256-GCM), auth status endpoint, custom HTTP client with TLS configuration.

#### `src/ui/server.go` (modified)
- Added `oidc *OIDCAuth` field to the `Server` struct
- Extended `NewServer()` signature to accept an optional `*OIDCAuth` parameter
- Added `registerAuthRoutes()` method that registers `/auth/login`, `/auth/callback`, `/auth/logout`, and `/api/v1/auth/status`
- Added `getAuthStatus()` handler that delegates to OIDC or returns `{"enabled": false}`
- OIDC middleware is conditionally wrapped around the router only when OIDC is configured

#### `src/cmd/main.go` (modified)
- Added 7 new optional configuration items: `OIDC_ISSUER_URL`, `OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET`, `OIDC_REDIRECT_URL`, `OIDC_SESSION_SECRET`, `OIDC_INSECURE_SKIP_VERIFY`, `OIDC_LOGOUT_URL`
- OIDC initialization block: when all three required variables (`OIDC_ISSUER_URL`, `OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET`) are set, the OIDC provider is initialized with a 30-second timeout context. When not all are set, a log message confirms unauthenticated access
- Updated `ui.NewServer()` call to pass the `*OIDCAuth` instance (or `nil`)

#### `src/static/index.html` (modified)
- Added `authFetch()` wrapper function that intercepts `401` responses from API calls and redirects the browser to `/auth/login`
- Replaced all `fetch()` calls to `/api/v1/*` endpoints with `authFetch()` (renovatejobs, discovery/start, renovate)
- Added `authInfo` state variable, loaded on mount via `GET /api/v1/auth/status`
- Added user info display in the header navigation: shows the user's name (or email) and a "Logout" button linking to `/auth/logout`, only visible when OIDC is enabled and the user is authenticated
- Theme toggle and user info are vertically stacked (theme toggle above, user/logout below) in the top-right area of the header

#### `Dockerfile` (modified)
- Added `COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt` to copy the CA certificate bundle from the Alpine builder stage into the `scratch` final image. This is required for Go's `net/http` client to verify TLS certificates when making outbound HTTPS requests (OIDC discovery, token exchange). Without this, all HTTPS connections fail with `x509: certificate signed by unknown authority`

#### `charts/renovate-operator/values.yaml` (modified)
Added new `auth.oidc` configuration block:

```yaml
auth:
  oidc:
    enabled: false
    issuerUrl: ""
    clientId: ""
    existingSecret: ""
    secretKey: "client-secret"
    sessionSecret: ""
    redirectUrl: ""
    insecureSkipVerify: false
    logoutUrl: ""
```

#### `charts/renovate-operator/templates/deployment.yaml` (modified)
Added conditional environment variable injection when `auth.oidc.enabled` is `true`:
- `OIDC_ISSUER_URL` — from `auth.oidc.issuerUrl`
- `OIDC_CLIENT_ID` — from `auth.oidc.clientId`
- `OIDC_CLIENT_SECRET` — from Kubernetes Secret reference (`auth.oidc.existingSecret` / `auth.oidc.secretKey`)
- `OIDC_SESSION_SECRET` — optional, from `auth.oidc.sessionSecret`
- `OIDC_REDIRECT_URL` — optional, from `auth.oidc.redirectUrl`
- `OIDC_INSECURE_SKIP_VERIFY` — optional, from `auth.oidc.insecureSkipVerify`
- `OIDC_LOGOUT_URL` — optional, from `auth.oidc.logoutUrl`

#### `src/go.mod` / `src/go.sum` (modified)
- Added direct dependency: `github.com/coreos/go-oidc/v3 v3.17.0`
- Added transitive dependency: `github.com/go-jose/go-jose/v4 v4.1.3`
- Promoted `golang.org/x/oauth2` from indirect to direct dependency (v0.30.0 → v0.35.0)

---

### Environment Variables Reference

| Variable | Required | Default | Description |
|---|---|---|---|
| `OIDC_ISSUER_URL` | Yes* | — | OIDC Issuer URL for provider discovery |
| `OIDC_CLIENT_ID` | Yes* | — | OAuth2 Client ID |
| `OIDC_CLIENT_SECRET` | Yes* | — | OAuth2 Client Secret |
| `OIDC_REDIRECT_URL` | No | — | OAuth2 callback URL (e.g. `https://renovate.example.com/auth/callback`) |
| `OIDC_SESSION_SECRET` | No | random | Secret for session cookie encryption. Set this to persist sessions across pod restarts |
| `OIDC_INSECURE_SKIP_VERIFY` | No | `false` | Skip TLS certificate verification for OIDC provider |
| `OIDC_LOGOUT_URL` | No | auto-discovered | OIDC end_session_endpoint URL for provider logout |

*Required only when OIDC should be enabled. When all three are set, OIDC is active. When any is missing, OIDC is disabled and access is unauthenticated.*

---

### Helm Values Example

```yaml
auth:
  oidc:
    enabled: true
    issuerUrl: "https://login.example.com"
    clientId: "renovate-operator"
    existingSecret: "renovate-oidc-secret"
    secretKey: "client-secret"
    sessionSecret: "my-stable-secret-for-cookie-encryption"
    redirectUrl: "https://renovate.example.com/auth/callback"
    logoutUrl: "https://login.example.com/logout"
```

---

### Backward Compatibility

- **Fully backward compatible** — no breaking changes
- Without OIDC environment variables: behavior is identical to before (unauthenticated access)
- Existing Helm deployments without `auth` values continue to work (all defaults are disabled/empty)
- Existing API clients are unaffected when OIDC is not configured
- The `/health` endpoint is never protected, ensuring Kubernetes probes always work

### DIFF
--- a/src/ui/oidc.go
+++ b/src/ui/oidc.go
@@ -36,8 +36,8 @@ type OIDCAuth struct {
 
 func NewOIDCAuth(ctx context.Context, cfg OIDCConfig, logger logr.Logger) (*OIDCAuth, error) {
 	transport := &http.Transport{
-		Proxy:                  http.ProxyFromEnvironment,
-		DialContext:            (&net.Dialer{Timeout: 10 * time.Second}).DialContext,
+		Proxy:                 http.ProxyFromEnvironment,
+		DialContext:           (&net.Dialer{Timeout: 10 * time.Second}).DialContext,
 		TLSHandshakeTimeout:   10 * time.Second,
 		ResponseHeaderTimeout: 10 * time.Second,
 	}
@@ -98,7 +98,7 @@ func NewOIDCAuth(ctx context.Context, cfg OIDCConfig, logger logr.Logger) (*OIDC
 }
 
 func (o *OIDCAuth) AuthMiddleware(next http.Handler) http.Handler {
-	return o.baseAuth.authMiddleware(next)
+	return o.authMiddleware(next)
 }
 
 func (o *OIDCAuth) HandleLogin(w http.ResponseWriter, r *http.Request) {
@@ -163,7 +163,7 @@ func (o *OIDCAuth) HandleCallback(w http.ResponseWriter, r *http.Request) {
 }
 
 func (o *OIDCAuth) HandleComplete(w http.ResponseWriter, r *http.Request) {
-	o.baseAuth.handleComplete(w, r)
+	o.handleComplete(w, r)
 }
 
 func (o *OIDCAuth) HandleLogout(w http.ResponseWriter, r *http.Request) {
@@ -182,5 +182,5 @@ func (o *OIDCAuth) HandleLogout(w http.ResponseWriter, r *http.Request) {
 }
 
 func (o *OIDCAuth) HandleAuthStatus(w http.ResponseWriter, r *http.Request) {
-	o.baseAuth.handleAuthStatus(w, r)
+	o.handleAuthStatus(w, r)
 }


mantoine96:
I'm really looking forward to this feature!

I have taken the liberty to extend it to integrate authorization in [this branch](https://github.com/mogenius/renovate-operator/compare/feat/oidc-support...mantoine96:renovate-operator:feat/oidc-support-with-authorization?expand=1), which I'm running in the meantime, but would love to upstream. I'll make a PR once this gets merged.

PR #145, opened 2026-02-23
feat: add Forgejo webhook sync controller

pat-s:
This PR adds automatic webhook management for Forgejo repos, removing the need to manually configure webhooks on each repository.

Forgejo lacks a "GitHub Apps" equivalent for automatic webhook installation, so webhooks must currently be added per-repo. This new controller automates that by querying the Forgejo API after each autodiscovery cycle, verifying admin permissions, and creating or removing webhooks on eligible repos.

Configuration lives under `spec.webhook.forgejo.sync` in the `RenovateJob` CRD, making it clear this is Forgejo-specific and entirely opt-in.

This PR is related to #144, i.e. the docs refer to the FJ-specific endpoint `/webhook/v1/forgejo`. For most parts however, this PR is working standalone and makes use of whatever endpoint the internal logic is set to provision.

### DIFF
--- a/docs/webhooks/forgejo.md
+++ b/docs/webhooks/forgejo.md
@@ -74,7 +74,6 @@ spec:
     forgejo:
       sync:
         enabled: true
-        forgejoURL: https://forgejo.example.com
         webhookURL: https://your-webhook-host/webhook/v1/forgejo
         topic: renovate
         tokenSecretRef:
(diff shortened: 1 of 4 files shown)


viceice:
i simply use a system webhook as workaround 😁

docs/webhooks/forgejo.md
```diff
@@ -22,44 +71,53 @@ spec:
       secretRef:
         name: renovate-webhook-token
         key: token
+    forgejo:
+      sync:
+        enabled: true
+        forgejoURL: https://forgejo.example.com
```
lukashankeln: Do we need the extra url field here? Or could we reference 'spec.provider.endpoint' ?

pat-s:
@beneiltis Thanks for the time! Addressed your review, good points!

> One more thing regarding the review timeline in general: we don't run Forgejo ourselves, which makes it hard for us to properly test this end-to-end. Combined with the size of the PR (~1300 lines), reviews take longer on our side — hope you understand.

Surely. Yet as a user one can't see whether the PR is actually "on the list" or just forgotten if there isn't any communication from the owners, so one has to ping occasionally ;)

> Please combine (rebase) your commits into 1 commit.

Any reason you don't just use squash merging? Rebasing and force-pushing into one commit during a PR flow makes it much harder to review. Just curious as I've never seen this policy in years in any repo and you guys seem to know what you're doing :)
I'll leave the latest commit like this until you checked on it and will then squash it.

@lukashankeln also addressed your point!

Issue #157, opened 2026-02-25
[Feature] Better support for CRDs when using ArgoCD


leneguer:
**Is your feature request related to a problem? Please describe.**
When deploying via ArgoCD, it is preferred to see the CRD diff when syncing.  Unfortunately, ArgoCD does not display hooks in the diff output, so any changes to the CRD are effectively invisible.

**Describe the solution you'd like**
There are two options that would easily resolve this:
1) rename the `charts/renovate-operator/crd` directory back to `charts/renovate-operator/crds`.  This would allow `helm template --include-crds` to work, and properly template the CRD on request.
2) add a values flag that would force templating of the CRD without the webhook.

leneguer:
On further thought, it occurred to me that the rename from `crds` to `crd` was intentionally made to bypass the default helm CRD handling.  Assuming this is the case, I opened https://github.com/mogenius/renovate-operator/pull/158 to fix this via method 2 above.

viceice:
ok. 🤔

PR #158, opened 2026-02-25
feat(helm): adding optional templating of CRDs

leneguer:
This is one option for fixing https://github.com/mogenius/renovate-operator/issues/157

### DIFF
--- a/charts/renovate-operator/values.yaml
+++ b/charts/renovate-operator/values.yaml
@@ -7,6 +7,7 @@ replicaCount: 1
 crd:
   # -- whether to install the CRD or not, set to false if you want to install the CRD separately
   install: true
+  mode: hook # optional: choose between "hook" and "template" install modes
   image:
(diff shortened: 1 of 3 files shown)


charts/renovate-operator/values.yaml
```diff
@@ -7,6 +7,7 @@ replicaCount: 1
 crd:
   # -- whether to install the CRD or not, set to false if you want to install the CRD separately
   install: true
+  mode: hook # optional: choose between "hook" and "configmap" install modes
```
lukashankeln: Wouldnt "hook" vs "template" be a better naming here?

leneguer: done.

leneguer:
> Thank you for your contribution 😄
> 
> Could you change the commit message to a semantic commit? e.g. `feat(helm): adding optional templating of crds`

fixed

PR #159, opened 2026-02-25
feat: set DNSPolicy in JobSpec

leneguer:
There are scenarios where the default DNSPolicy of "ClusterFirst" won't work.  In my case, I need renovate to use the "Default" DNSPolicy to bypass CoreDNS and the rewrites I'm doing there to MitM docker.io for Docker-in-Docker workloads.

This maintains the default "ClusterFirst" behavior, but adds the ability to specify an alternate DNSPolicy if required.


lukashankeln (approved):

PR #170, opened 2026-02-27
fix: add TTLSecondsAfterFinished to discovery job spec

acuntex:
Discovery runs create Kubernetes `Job` resources that currently remain in the cluster after completion.

This PR sets `ttlSecondsAfterFinished` on the discovery job spec (using the existing `getJobTTLSecondsAfterFinished()` helper), so finished discovery jobs are cleaned up automatically.

This makes the behavior consistent with the regular Renovate jobs, which already use the same TTL mechanism.

As a result, namespaces stay clean and we avoid accumulating completed jobs and pods over time, especially in setups with frequent discovery runs.


lukashankeln (approved):

PR #174, opened 2026-02-27
feat: Add httproute fields

pierreyves-lebrun:
### Problem

`Httproute` is not fully configurable so it's currently not possible to specific fields like `group`, `kind` or `weight`.

### Solution

Add those fields to `values.yaml` with default values and let the user override them when necessary.


lukashankeln (approved):

PR #178, opened 2026-03-01
feat: Add authorization support to OIDC authentication

mantoine96:
This PR adds authorization support to the OIDC authentication method. 

This enables administrators to define who can see which jobs based on:
* `defaultAllowedGroups`: groups which will be allowed to see and interact with all Renovate Jobs
* `allowedGroups` on the Renovatejob: group which will be allowed to see and interact with the Renovate Job

Additionally, given some (for example my job's) OIDC environments, we can easily have hundreds of groups, so I've added some filtering for the groups so only those relevant can be populated in Renovate Operator.

This change has not been implemented for GitHub for now, to keep the changes minimal. Let me know if this is something you would like though, and I'll be glad to add it.

### DIFF
--- a/src/ui/renovateController.go
+++ b/src/ui/renovateController.go
@@ -30,6 +30,178 @@ type ExecutionOptions struct {
 	Debug bool `json:"debug,omitempty"`
 }
 
+// filterRenovateJobsByGroups filters jobs based on user groups and job's allowedGroups.
+// Authorization rules:
+// - When auth is disabled (authEnabled == false): all jobs visible
+// - When auth is enabled (authEnabled == true):
+//   - Jobs without allowedGroups use defaultAllowedGroups (from operator config)
+//   - If defaultAllowedGroups is also empty, job is HIDDEN (secure by default)
+//   - Jobs with allowedGroups shown only if user has at least one matching group
+//   - Users without groups see no jobs
+//   - If session is nil (edge case/bug), return empty list for security
+func filterRenovateJobsByGroups(jobs []api.RenovateJob, authEnabled bool, session *sessionData, defaultAllowedGroups []string) []api.RenovateJob {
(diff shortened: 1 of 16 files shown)


src/ui/renovateController.go
```diff
@@ -24,6 +24,177 @@ type RenovateJobInfo struct {
...
+	// Audit log the authorization decision
+	if authorized {
+		s.logger.V(1).Info("Authorization granted",
+			"user", session.Email,
+			"user_groups", session.Groups,
+			"resource", jobName,
+			"namespace", namespace,
+			"allowed_groups", effectiveAllowedGroups,
+			"path", r.URL.Path,
+			"method", r.Method)
```
ruedigerp: Here the IP should also be logged ("remote_addr", r.RemoteAddr) just like with the authorization denied.

mantoine96: Hey @ruedigerp ! Thank you! Done

mantoine96:
Thank you @ruedigerp for your review! 

Should we do something regarding @jduepmeier 's issue? Maybe in a subsequent PR?

beneiltis:
Hey Matthieu, thanks for this — really solid work.

Two questions before we merge (today ;-) ):

1. Spaces in group names — the validation regex rejects spaces, but Azure AD and other providers often return groups with spaces (e.g. "All Users", "Team Alpha"). Those get silently dropped. Was that intentional?

2. GitHub auth — with GitHub OAuth the session always has empty groups, so anyone using GitHub auth + allowedGroups would see zero jobs without any hint why. I think we should log a warning until GitHub auth is active too. what do you think?

Sorry it takes so long with the big PRs but they are not easy to review. But thanks a lot for your contribution :-)

beneiltis:
Dont forget to combine all changes in one commit. Thx :-)

PR #212, opened 2026-03-16
fix: add mutex to prevent crashing with shared cron schedules

### LINKED ISSUE #211: Concurrent map writes panic when multiple cron schedules fire simultaneously
**Describe the bug**
When two or more `RenovateJob` resources share the same cron schedule (e.g. `0 6-16 * * 1-5`), the operator panics with a fatal `concurrent map writes` error the moment those schedules fire at the same time.

The root cause is that `healthcheck` in `src/health/health.go` has no synchronisation. `SetSchedulerHealth` (and `SetExecutorHealth` / `GetHealth`) read and write the shared `Scheduler` and `Executor` maps without a mutex. When the cron library fires multiple jobs concurrently, each goroutine calls `SetSchedulerHealth` in `scheduler.execute()`, and the concurrent map writes crash the process.

### Stack trace
```
fatal error: concurrent map writes

goroutine 1665 [running]:
internal/runtime/maps.fatal({0x16bc4e4?, 0x34c1c?})
runtime/panic.go:1181 +0x20
renovate-operator/health.(*healthcheck).SetSchedulerHealth(0x33a180f02228, 0x3212c?)
renovate-operator/health/health.go:77 +0x128
renovate-operator/scheduler.(*scheduler).execute.func1()
renovate-operator/scheduler/scheduler.go:141 +0xc0
```

### Impact

The operator crashes every time multiple schedules align, then restarts with a fresh in-memory scheduler that re-calculates its next fire time — often missing the window entirely. In practice this means cron-triggered jobs frequently don't run.

### Suggested fix

Add a `sync.RWMutex` to the `healthcheck` struct and acquire it in `SetSchedulerHealth`, `SetExecutorHealth`, and `GetHealth`.

**To Reproduce**
Have more than one cron for the exact same time and watch the sadness.

**Expected behavior**
Ideally, less sadness :)

**Screenshots & Logs**

 - Kubernetes Version [e.g. v1.34.2] 1.35.0
 - Version [e.g. 1.8.0] 3.1.0

**Additional context**

tico24:
Closes #211

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -43,12 +43,10 @@ jobs:
         go-version: '>=1.25.4'
 
     - name: lint and test
-      env:
-        CGO_ENABLED: 1
       run: |
         cd src
         go mod download
-        go test -race ./... -v
+        go test ./... -v
 
   lint:
     needs: [build]
--- a/src/health/health_test.go
+++ b/src/health/health_test.go
@@ -77,6 +77,9 @@ func TestConcurrentExecutorHealthUpdates(t *testing.T) {
 	wg.Wait()
 }
 
+// TestConcurrentReadWriteHealth should ideally be run with -race to detect data races,
+// but that requires CGO and a C compiler on the runner, and is not present on the self-hosted
+// runners used by the CI.
 func TestConcurrentReadWriteHealth(t *testing.T) {
 	t.Parallel()
 


.github/workflows/go-test.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
       run: |
         cd src
         go mod download
-        go test ./... -v
+        go test -race ./... -v
```
tico24: I don't know if you have a C compiler on your self-hosted runners. I'll add CGO_ENABLED and if it fails again I'll remove the race tests I guess?

tico24: There is indeed no C Compiler on the self-hosted runner so I have reverted the race flag and added a comment to explain that ideally it should be tested with it. Best I can do given the circumstances. The test is still useful, just not as useful as it could be.

PR #213, opened 2026-03-16
fix: add mutex to healthcheck to prevent concurrent map writes

mantoine96:
The operator panics at runtime with:

```
fatal error: concurrent map writes

goroutine 234 [running]:
internal/runtime/maps.fatal({0x1dc4e81?, 0x104e5e5ddee0?})
        runtime/panic.go:1181 +0x18
renovate-operator/health.(*healthcheck).SetSchedulerHealth(0x104e5e4e0348, 0x425b2c?)
        renovate-operator/health/health.go:77 +0x1a5
renovate-operator/scheduler.(*scheduler).execute.func1()
        renovate-operator/scheduler/scheduler.go:141 +0xe2
github.com/netresearch/go-cron.FuncJob.Run(0x2f9b280?)
        github.com/netresearch/go-cron@v0.13.1/cron.go:528 +0x12
github.com/netresearch/go-cron.(*Cron).startJobWithExecution.func1.1(0x428513?, {0x1f39320?, 0x104e60d482c0?}, {0x1f4f3b0?, 0x104e60d0ec30?})
        github.com/netresearch/go-cron@v0.13.1/cron.go:1306 +0x8d
github.com/netresearch/go-cron.(*Cron).startJobWithExecution.func1()
        github.com/netresearch/go-cron@v0.13.1/cron.go:1308 +0x18b
created by github.com/netresearch/go-cron.(*Cron).startJobWithExecution in goroutine 136
        github.com/netresearch/go-cron@v0.13.1/cron.go:1290 +0x20c
```

  The healthcheck struct in health/health.go holds an *ApplicationHealth containing map[string]SingleSchedulerHealth and map[string]SingleExecutorHealth. 

Multiple cron goroutines call SetSchedulerHealth concurrently (via scheduler.execute()), all writing to the same underlying map with no synchronization.

The same race exists between SetExecutorHealth calls and between GetHealth reads (HTTP handler) and any concurrent write.

In this PR we: 
  - Add a sync.RWMutex to the healthcheck struct
  - SetSchedulerHealth / SetExecutorHealth: acquire exclusive Lock around the entire method body
  - GetHealth: acquire RLock and return a deep copy of ApplicationHealth using maps.Clone, so no map reference escapes the lock — preventing races when the HTTP handler iterates the maps via json.Encode

### DIFF
--- a/src/health/health.go
+++ b/src/health/health.go
@@ -1,6 +1,10 @@
 package health
 
-import "time"
+import (
+	"maps"
+	"sync"
+	"time"
+)
 
 type SchedulerHealth struct {
 	Running   bool                             `json:"running"`
@@ -34,6 +38,7 @@ type HealthCheck interface {
 	SetSchedulerHealth(func(health *SchedulerHealth) *SchedulerHealth)
 }
 type healthcheck struct {
+	mu     sync.RWMutex
 	health *ApplicationHealth
 }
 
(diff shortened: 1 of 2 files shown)


mantoine96:
Oh I see #212 was created to fix this. Will close this then.

Issue #181, opened 2026-03-02
[Feature] Persist logs of finished Renovate jobs (independent of Kubernetes Job TTL)


acuntex:
Hi!

First of all, thanks for the great work on the Renovate Operator!

## Current Behavior

When clicking on Logs in the Web UI for a Renovate job, the operator fetches the logs directly from the associated Kubernetes Pod.

This works perfectly while the job is still running.

However, if `jobTTLSecondsAfterFinished` is configured, Kubernetes deletes the Job (and therefore the Pod) after the TTL expires. As a result:

- The Pod is deleted
- The logs are no longer available
- The Web UI cannot display logs for previous runs

## Problem

In environments where `jobTTLSecondsAfterFinished` is enabled (which is a reasonable cleanup strategy), historical logs of Renovate runs are lost shortly after completion.

This makes troubleshooting or reviewing past runs difficult, especially if issues are discovered later.

## Proposed Enhancement

It would be great if the Renovate Operator could persist the logs of the last job runs independently of the Kubernetes Job lifecycle.

For example:

- After a job completes, the operator could retrieve and store the logs.
- Logs could be persisted in a lightweight storage backend (e.g. SQLite).
- The Web UI would then fetch logs from this internal storage instead of directly from the Pod when the Pod no longer exists.

## Possible Implementation Ideas

- Store logs per Renovate CR (e.g., last run, maybe with timestamps and status).
- Keep only the most recent run to avoid unbounded storage growth.
- Make log persistence configurable (enabled/disabled/storage path).
- Optionally support a configurable retention policy (e.g., number of runs or time-based retention).

## Benefits

- Logs remain available even when jobTTLSecondsAfterFinished is enabled.
- Better observability and debugging of past Renovate runs.
- Cleaner Kubernetes clusters without sacrificing log visibility.

If this idea aligns with the project direction, I’d be happy to help further clarify requirements or discuss possible approaches.

Thanks!

acuntex:
That makes sense regarding multiple replicas.

However, I'm wondering whether we need an additional dependency for the common case of running a single operator replica. I suspect many users run `replicas=1`, where log retention could be implemented much more lightweight.

Would you be open to a staged approach?

- **Minimal/default**: `logStorage.mode=memory` (or disabled by default)
Keep "last run" logs in an in-memory ring buffer, bounded by `maxRuns`/`maxBytes`/`maxLines`.
[...]

PR #225, opened 2026-03-20
fix: Store sessions server-side to prevent cookie size overflow

mantoine96:
In #178 we introduced authorization and started storing user information, including group memberships, in a browser cookie.

However, users with many groups can exceed the 4096-byte cookie size limit, causing the browser to truncate it. This breaks authentication entirely, with the user hitting the authentication loop prevention error.

Taking inspiration from oauth2-proxy and similar projects, sessions are now stored server-side. The cookie only contains an AES-GCM encrypted session ID, keeping it well within browser limits.

This PR adds:
  * A server-side session store with two backends: in-memory (single replica) and Redis (multi-replica deployments)
  * A background sweep mechanism to clean up expired sessions
  * An optional Redis subchart dependency in the Helm chart
  * Encryption hardening and deduplication of the AES-GCM helpers
  * Updated auth documentation (docs/auth.md)
  * Tests for the session store across both Redis (using [miniredis](https://github.com/alicebob/miniredis)) and the in-memory implementation.

Upgrade notes: 
* Existing sessions will be invalidated on upgrade. Users will need to re-authenticate. 
* Operators opting for Redis must configure the new redis.* Helm values. 
* On `helm upgrade` of an existing application, redis.password will need to be explicitly set (Helm will not be able generate it itself).


charts/renovate-operator/Chart.yaml
```diff
@@ -14,3 +14,8 @@ maintainers:
   - email: lukas@mogenius.com
     name: Lukas Hankeln @mogenius
     url: https://mogenius.com
+dependencies:
+  - name: redis
+    version: "25.3.5"
+    repository: oci://registry-1.docker.io/bitnamicharts
```
viceice: bitnami isn't a good solution. the used images are no longer open source

i think valkey chart is a better solution

- <https://github.com/valkey-io/valkey-helm>

src/ui/redis_store.go
```diff
@@ -0,0 +1,100 @@
...
+// The redisURL should include credentials and DB if needed
+// (e.g. "redis://:password@redis:6379/0").
+// The encryptionKey is used to encrypt session data at rest.
+func NewRedisSessionStore(redisURL string, encryptionKey [32]byte) (SessionStore, error) {
+	opts, err := redis.ParseURL(redisURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid Redis URL: %w", err)
+	}
+
+	client := redis.NewClient(opts)
```
viceice: does it support redis sentinel or cluster?

charts/renovate-operator/Chart.yaml
lukashankeln: 100%

Valkey also has its own library, i think we should also switch to that one in the process

src/cmd/main.go
```diff
@@ -262,14 +282,50 @@ func main() {
 
 	cronManager := scheduler.NewScheduler(ctrl.Log.WithName("scheduler"), health)
 
-	// Initialize authentication provider (OIDC or GitHub OAuth)
-	var authProvider ui.AuthProvider
+	// Determine the session secret for the active auth provider so we can
```
lukashankeln: The complete auth init is getting pretty long here, can we move that to separate function?

src/cmd/main.go
```diff
@@ -262,14 +282,50 @@ func main() {
...
+	sessionStore, storeType, storeErr := ui.NewSessionStore(ui.RedisConfig{
+		URL:      config.GetValue("REDIS_URL"),
+		Host:     config.GetValue("REDIS_HOST"),
+		Port:     config.GetValue("REDIS_PORT"),
+		Password: config.GetValue("REDIS_PASSWORD"),
+	}, storeKey)
+	assert.NoError(storeErr, "failed to initialize session store")
+	ctrl.Log.WithName("auth").Info("Using session store", "type", storeType)
+	if storeType == "memory" && leaderElectionID != "" {
+		ctrl.Log.WithName("auth").Info("WARNING: in-memory session store with multiple replicas will cause sessions to break across pods; configure Redis for multi-replica deployments")
```
lukashankeln: would it be possible to only use the server side session if redis/valkey is enabled?

otherwise this would be a breaking change

src/ui/redis_store.go
lukashankeln, line 1: We also have #181 that might profit from a valkey connection. 
Maybe we could make the redis/valkey store a little bit more generic, e.g. with a Put(... keys) and Get(...keys) functions. 
That would move the `sessions:` key to the session store itself and the redis/valkey store could be reused for other cases

lukashankeln (changes requested):

charts/renovate-operator/Chart.yaml
mantoine96: Done!

src/cmd/main.go
mantoine96: Done!

src/ui/redis_store.go
mantoine96: Agreed, done

src/cmd/main.go
mantoine96: Yes agreed, done!

src/ui/redis_store.go
mantoine96: Now with Valkey, thanks to `valkey.ParseUrl` (see [here](https://github.com/valkey-io/valkey-go#valkey-url)) we should be able to support pretty much any configuration of Valkey; standalone or cluster.

lukashankeln:
@ruedigerp & @beneiltis could you take a look at this?
For me its looking good, but as it is auth related, six eyes are better than two 😆

beneiltis (approved):
Looks good to me. Took some time because its a big boy :-)

PR #232, opened 2026-03-27
feat(chart): add additionalRules field for httproutes

wittdennis:
Allows for arbitrary defining of HttpRouteRules which can be useful for certain use cases

### DIFF
--- a/charts/renovate-operator/values.yaml
+++ b/charts/renovate-operator/values.yaml
@@ -97,11 +97,16 @@ route:
   # -- parentRefs to place on the HTTPRoute
   parentRefs: []
   # -- group to place on the HTTPRoute
-  group: ''
+  group: ""
   # -- kind to place on the HTTPRoute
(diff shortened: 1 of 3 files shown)


charts/renovate-operator/values.yaml
```diff
@@ -97,11 +97,16 @@ route:
...
   parentRefs: []
   # -- group to place on the HTTPRoute
-  group: ''
+  group: ""
   # -- kind to place on the HTTPRoute
   kind: Service
   # -- weight to place on the HTTPRoute
   weight: 1
+  # -- Any custom rules you want to specify
+  additionalRules: {}
```
lukashankeln: Shouldnt this be a list? e.g.: '[]'

wittdennis: That is technically correct. The best kind of correct! 

Updated!

PR #237, opened 2026-03-31
feat: repo-level access filtering and write permission enforcement in dashboard

pat-s:
## Summary

- Adds repo-level project filtering in the dashboard UI using the user's OAuth access token to query the git platform API (Forgejo/Gitea and GitHub supported)
- Repos where the user only has read access are collapsed behind a toggle, rendered dimmed with a "read-only" badge, and cannot trigger renovate runs
- The trigger API (`/api/v1/renovate`) also enforces write permission server-side, returning 403 for read-only repos
- Introduces a `RepoCache` interface with an in-memory implementation (5 min TTL); designed to be swappable for Valkey/Redis
- Stores the OIDC access token in the session (GitHub OAuth already did this)

### Changes

| File | Change |
|---|---|
| `src/ui/repo_access.go` | **New** — `RepoCache` interface, `MemoryRepoCache`, platform repo fetchers with permission tracking, `filterProjectsByAccess`, `canUserWriteRepo` |
| `src/ui/repo_access_test.go` | **New** — 11 tests covering cache, permission-aware fetchers, filtering, write checks, and edge cases |
| `src/ui/oidc.go` | Store `oauth2Token.AccessToken` in session during OIDC callback |
| `src/ui/renovateController.go` | New `UIProjectStatus` type with `canWrite` field; annotate projects with write permissions; enforce write access on trigger endpoint |
| `src/ui/server.go` | Add `repoCache` field to `Server` |
| `src/cmd/main.go` | Create `MemoryRepoCache` and pass to `NewServer` |
| `src/static/index.html` | Split project list into writable (shown) and read-only (collapsed toggle); disable trigger for read-only |


pat-s:
@lukashankeln bump :)

pat-s:
@lukashankeln This is important to us. Would be great to move this forward!

robinschneider:
This PR looks very close to the use case I have as well: making the dashboard safe to expose publicly while preventing users without write access from triggering Renovate runs.

For my setup, the key requirement is that authorization is enforced on the API/server side too, not only reflected in the UI. It would also be useful if the behavior could work with OIDC role/group mappings where repository-level write permissions are not available or not desired.

I would be very interested in seeing this move forward, because it solves a real operational problem for public dashboards.

In my use case the dashboard would be public and all interactions such as discovery and renovate runs for users that authenticated with a 3rd party OIDC provider only.

wittdennis:
Thank you for your PR. We gave this much thought but ultimately decided against this implementation. 

But we can also see that this is a feature that has demand and will work on a implementation that we feel aligns more with the project. 

To take part in the discussion you can follow #380

pat-s:
Uff, that's a bummer. :/

Note that this logic is [in use since several months](https://renovate.codefloe.com/) meanwhile and works without any issues, using the internal RBAC logic of the respective forge.

Custom OIDC roles can surely be an alternative but arguably forge-internal RBAC should always be supported?

It would have also been great to get a bit more detail/honest reply on why this implementation was rejected than a "aligns more with the project" which is super unspecific boilerplate.

PR #240, opened 2026-04-02
fix: make job name sanatization conform to k8s naming rules

### LINKED ISSUE #236: Job name
**Describe the bug**
We use Azure for our git repos. We use projects and project groups with spaces and other special characters. When the operator tries to spawn k8s jobs, the creation errors due to invalid characters in the job name.

**To Reproduce**
1. Setup renovate-operator connected to an Azure Devops environment with project names that don't follow k8s naming convention requirements.
2. Attempt to trigger a project with invalid names
3. See error

**Expected behavior**
The operator should translate project names into acceptable job names potentially replacing invalid characters with '-'

**Screenshots & Logs**

 - Kubernetes Version Latest
 - Version 3.10.0

wittdennis:
Fixes #236

### DIFF
--- a/src/internal/utils/jobNames_test.go
+++ b/src/internal/utils/jobNames_test.go
@@ -61,14 +61,14 @@ func TestRenovateJob_JobNames(t *testing.T) {
 			name:              "complex project name",
 			jobName:           "renovate",
 			project:           "Org/.github",
-			expectedExecutor:  "renovate-org--github-3ff30a25",
+			expectedExecutor:  "renovate-org-github-cb19908c",
 			expectedDiscovery: "renovate-discovery-df3ed160",
 		},
 		{
 			name:              "long project name",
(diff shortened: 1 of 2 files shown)


src/internal/utils/jobNames_test.go
```diff
@@ -61,14 +61,14 @@ func TestRenovateJob_JobNames(t *testing.T) {
 			name:              "complex project name",
 			jobName:           "renovate",
 			project:           "Org/.github",
-			expectedExecutor:  "renovate-org--github-3ff30a25",
```
lukashankeln: This is a breaking change...

Depending on your configuration this might result in Jobs not getting deleted

wittdennis: Fair enough. I marked the change as breaking. Would make sense then to bundle with #241

PR #247, opened 2026-04-02
feat(ui): add settings to hide projects by status

tibuntu:
## Summary
- Adds per-status toggles in the settings menu to hide projects by their renovate result status (Onboarding Closed, Disabled, No Config)
- Each toggle shows the count of matching projects
- Preferences are persisted in localStorage across page reloads

## Test plan
- [ ] Verify toggles appear in the gear icon settings dropdown under "Hide Projects"
- [ ] Verify checking a toggle hides only projects with that specific status
- [ ] Verify multiple statuses can be hidden simultaneously
- [ ] Verify preferences persist after page reload
- [ ] Verify counts update correctly as projects change status

### DIFF
--- a/src/static/index.html
+++ b/src/static/index.html
@@ -1128,7 +1128,35 @@
           key: "status",
           direction: "asc",
         });
-        const sortedProjects = useMemo(() => getSortedProjects(job.projects, sortConfig), [job.projects, sortConfig]);
+        const STATUS_FILTERS = ["Onboarding Closed", "Disabled", "No Config"];
+        const [hiddenStatuses, setHiddenStatuses] = useState(() => {
+          try {
+            const stored = JSON.parse(localStorage.getItem('hiddenStatuses') || '[]');
+            return Array.isArray(stored) ? stored : [];
+          } catch {
+            return [];
+          }
+        });
+        const toggleStatusFilter = (status) => {
+          setHiddenStatuses((prev) => {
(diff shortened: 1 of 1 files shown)


src/static/index.html
```diff
@@ -1415,6 +1436,19 @@ <h3 className="text-sm font-semibold text-gray-900 dark:text-slate-100 mb-3">Exe
 
             {open && (
               <div className="border-t border-gray-200 dark:border-slate-700 bg-gray-50 dark:bg-slate-900/50">
+                {onboardingCount > 0 && (
```
lukashankeln: I think we should place this in the settings menue

src/static/index.html
```diff
@@ -1128,7 +1128,28 @@ <h3 className="text-lg font-semibold text-gray-700 dark:text-slate-300 mb-2">
           key: "status",
           direction: "asc",
         });
-        const sortedProjects = useMemo(() => getSortedProjects(job.projects, sortConfig), [job.projects, sortConfig]);
+        const [hideOnboarding, setHideOnboarding] = useState(() => {
+          try {
+            return localStorage.getItem('hideOnboarding') === 'true';
```
lukashankeln: why only onboarding disabled? could this be more generic so that we could add settings for No Config, Disabled, etc as well?

tibuntu: Valid point, in my current renovate-operator setup, I only have a single repo with the onboarding disabled status, thus I didn't consider other situations.

PR #256, opened 2026-04-04
feat: Add scratch volume and base dir

pierreyves-lebrun:
# Summary

## Motivation

- Renovate’s working directory was implicitly tied to `/tmp`: hardcoded `RENOVATE_BASE_DIR`, discovery paths, and the emptyDir mount path did not stay in sync when users wanted to use a different directory
- The built-in scratch emptyDir had no way to set `sizeLimit` / `medium` or to express ephemeral-storage request/limit

### `spec.renovateBaseDir`

- This field is optional but acts as the ingle source of truth for **`RENOVATE_BASE_DIR`** and the scratch emptyDir **mount path** (default **`/tmp`** when unset)
- Any **`RENOVATE_BASE_DIR` entry in `extraEnv` is ignored** so the path cannot drift from the volume mount

### Discovery job

- Replaced hardcoded `/tmp/repos.json` and `/tmp/logs.json` with paths under **`$RENOVATE_BASE_DIR`**

### `spec.scratchVolume` (`RenovateJobScratchVolume`)

- Optional tuning for the **built-in emptyDir** that backs `RENOVATE_BASE_DIR`:
  - **`sizeLimit`** → `emptyDir.sizeLimit`
  - **`medium`** → `emptyDir.medium` (e.g. `Memory` for tmpfs)

### Volume naming

- The pod volume and volumeMount name changed from **`tmp`** to **`scratch`** (`builtInScratchVolumeName`).
- If you reference this volume from **`extraVolumeMounts`**, use **`name: scratch`** (not `tmp`).

### CRD

- Helm CRD manifest updated: **`renovateBaseDir`**, **`scratchVolume`**

### Docs & tests

- **`docs/extra-volumes.md`**: default scratch volume, `renovateBaseDir`, and `scratchVolume` described.
- **`internal/renovate/jobDefinitions_test.go`**: coverage for `renovateBaseDir`, ignoring `RENOVATE_BASE_DIR` in `extraEnv`, and `scratchVolume` (emptyDir + ephemeral resources).

## Example

```yaml
spec:
  renovateBaseDir: /var/renovate
  scratchVolume:
    medium: Memory
    sizeLimit: 4Gi
```

### DIFF
--- a/src/internal/renovate/jobDefinitions.go
+++ b/src/internal/renovate/jobDefinitions.go
@@ -15,6 +15,8 @@ import (
 	"k8s.io/utils/ptr"
 )
 
+const defaultRenovateBaseDir = "/tmp"
+
 // create job spec for a discovery job
 func newDiscoveryJob(job *api.RenovateJob) *batchv1.Job {
 	predefinedEnvVars := getDefaultEnvVars(job)
@@ -48,21 +50,7 @@ func newDiscoveryJob(job *api.RenovateJob) *batchv1.Job {
 		envFromSecrets = append(envFromSecrets, job.Spec.ExtraEnvFrom...)
 	}
 
-	volumes := []v1.Volume{
-		{
-			Name: "tmp",
-			VolumeSource: v1.VolumeSource{
-				EmptyDir: &v1.EmptyDirVolumeSource{},
-			},
-		},
-	}
-
-	volumeMounts := []v1.VolumeMount{
-		{
-			Name:      "tmp",
-			MountPath: "/tmp",
-		},
-	}
+	containerEnv := mergeEnvVars(job.Spec.ExtraEnv, predefinedEnvVars)
 
 	batchJob := &batchv1.Job{
 		Spec: batchv1.JobSpec{
@@ -78,12 +66,12 @@ func newDiscoveryJob(job *api.RenovateJob) *batchv1.Job {
 						{
 							Name:            "discovery",
 							Command:         []string{"/bin/sh", "-c"},
-							Args:            []string{"renovate --autodiscover --write-discovered-repos /tmp/repos.json >> /tmp/logs.json 2>&1 && cat /tmp/repos.json || cat /tmp/logs.json"},
+							Args:            []string{`renovate --autodiscover --write-discovered-repos "$RENOVATE_BASE_DIR/repos.json" >> "$RENOVATE_BASE_DIR/logs.json" 2>&1 && cat "$RENOVATE_BASE_DIR/repos.json" || cat "$RENOVATE_BASE_DIR/logs.json"`},
 							Image:           job.Spec.Image,
-							Env:             mergeEnvVars(job.Spec.ExtraEnv, predefinedEnvVars),
+							Env:             containerEnv,
 							EnvFrom:         envFromSecrets,
 							Resources:       job.Spec.Resources,
-							VolumeMounts:    append(volumeMounts, job.Spec.ExtraVolumeMounts...),
+							VolumeMounts:    job.Spec.ExtraVolumeMounts,
 							SecurityContext: getContainerSecurityContext(job.Spec),
 						},
 					},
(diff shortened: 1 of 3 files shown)


src/internal/renovate/jobDefinitions.go
```diff
@@ -348,6 +400,27 @@ func getDNSPolicy(spec api.RenovateJobSpec) v1.DNSPolicy {
...
+func defaultRenovateBaseDir(spec api.RenovateJobSpec) string {
+	if spec.RenovateBaseDir != "" {
+		return spec.RenovateBaseDir
+	}
+	return "/tmp"
+}
+
+// withCanonicalRenovateBaseDir drops any RENOVATE_BASE_DIR from env (e.g. extraEnv) and appends
+// the value from spec.renovateBaseDir so the scratch mount and env stay a single source of truth.
+func withCanonicalRenovateBaseDir(env []v1.EnvVar, spec api.RenovateJobSpec) []v1.EnvVar {
```
lukashankeln: The hard overwrite of the `RENOVATE_BASE_DIR` after the merge of the env vars, removes the ability from the user to set that variable himself.

That could be breaking for existing Setups, I think we should set the variable in the default env vars and let the user allow to overwrite it

src/internal/renovate/jobDefinitions.go
```diff
@@ -198,6 +193,63 @@ func newRenovateJob(job *api.RenovateJob, project string) *batchv1.Job {
...
+	if spec.ScratchVolume.Medium != "" {
+		src.Medium = spec.ScratchVolume.Medium
+	}
+	if spec.ScratchVolume.SizeLimit != nil {
+		src.SizeLimit = spec.ScratchVolume.SizeLimit
+	}
+	return src
+}
+
+func mergeResourcesWithScratchVolume(base v1.ResourceRequirements, scratch *api.RenovateJobScratchVolume) v1.ResourceRequirements {
```
lukashankeln: The ephemeral-storage limits can already be set through `spec.resources`, what is the benefit of adding a separate entry for this?

lukashankeln (changes requested):

src/internal/renovate/jobDefinitions.go
pierreyves-lebrun: Fair enough, I removed those fields

src/internal/renovate/jobDefinitions.go
pierreyves-lebrun: I thought about that and I think that approach can be problematic. Consider the following example:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-config
  namespace: renovate
data:
  RENOVATE_BASE_DIR: /var/misaligned-path
---
apiVersion: renovate-operator.mogenius.com/v1alpha1
kind: RenovateJob
metadata:
  name: misaligned-scratch-example
  namespace: renovate
spec:
  schedule: "0 2 * * *"
  image: renovate/renovate:43
  provider:
    name: github
  parallelism: 1

  # renovateBaseDir omitted → operator mounts scratch at /tmp (default)

  scratchVolume:
    medium: Memory
    sizeLimit: 4Gi

  extraEnvFrom:
    - configMapRef:
        name: my-config # contains RENOVATE_BASE_DIR set to anything but /tmp → does not match scratch mount
```

Inferring the effective `RENOVATE_BASE_DIR` from ConfigMaps or Secrets during reconciliation is awkward: the controller would need to resolve those references and keep the scratch mount path in sync, which adds complexity and failure modes. I’d rather treat this as a breaking change and make the contract explicit in the API (e.g. require renovateBaseDir to match how scratch is mounted).

lukashankeln: Yes, but in that case i would say it is in the reponsibility of the user to mount a volume for his base-dir via the volumeMounts feature.
(it works the same at the moment)

I would not go that far as to the operator reading the env vars to check which volume to mount.

The current state does the same, it always mounts /tmp and if the user provides alternatives those will be used.

-> If a user with the current setup is mounting a pvc and setting the baseDir via his own env vars, this would be a breaking change.

(Also mounting a pvc might be a reasonable thing for chaching, so we should leave the user with this option)

pierreyves-lebrun: I am fine with the idea that mounting a volume for base dir is the user responsibility but in that case I would suggest we remove the default emptyDir on /tmp as even without this PR there is the problem that you can't infer RENOVATE_BASE_DIR as it can be defined via extraEnvFrom.

You can try to replicate this issue with the latest version of the operator by defining a volume for tmp, it will end up with a duplicated tmp volume entry due to https://github.com/mogenius/renovate-operator/blob/main/src/internal/renovate/jobDefinitions.go#L51

lukashankeln:
Closing in Favor of #284

Issue #275, opened 2026-04-09
badges as filters - PR badge only PoC


vtmocanu:
Based on discussion here
https://github.com/mogenius/renovate-operator/issues/259#issuecomment-4206185592

My proposal would be to be able to click on badges and they will act as filters, show only relevant Jobs/repos.
When a badge is clicked/toggled it must be made clear for user so (s)he knows (s)he can/should reclick it to get out of the filter.

I'll sketch a PoC PR for PRs and if you like it we can replicate for all badges.

PR #276, opened 2026-04-09
feat(ui): make Open PRs header badge a clickable filter toggle

vtmocanu:
## Summary

Makes the "Open PRs" header badge clickable as a toggle filter, hiding projects (and entire JobCards) that don't have any currently-open PRs. This implements the Open PRs portion of #275 and is ready to merge as-is. Follow-up work for the other 5 header badges is intentionally out of scope — see "Follow-ups" below.

Related: #259 / #270 (v4.2.0 badges + filter).

## Problem

The v4.2.0 Web UI shows PR-count badges in the header and a per-card "Without PRs" dropdown filter. The dropdown is awkward for the common case of "just show me the repos with open PRs" — users who spot the badge want to click it directly. #275 proposes clickable badges as filters.

## Changes

- **`StatBadge.js`**: gains optional `onClick`, `active`, `activeClass`, `title` props. When `onClick` is provided, the badge renders as a keyboard-accessible button (`role=button`, `tabIndex=0`, Enter/Space handlers, `aria-pressed`). Fully backward-compatible — the 5 non-clickable callers render identically.
- **`index.html`**:
  - New top-level `headerFilter` state in `App`, persisted to `localStorage`.
  - "Open PRs" header badge wired with `onClick={() => toggleHeaderFilter('openPRs')}`.
  - Active visual is a background fill + border (`bg-blue-50 dark:bg-blue-900/30 border-blue-500 dark:border-blue-400`) so it doesn't stack with the browser focus ring and is obvious in both light and dark mode.
  - `JobCard` accepts a `headerFilter` prop and applies the matching predicate in its `sortedProjects` useMemo. Predicate matches `stats.openPRs` exactly (`created + updated + unchanged`, excluding automerged).
  - New `visibleJobs` useMemo at the `App` level hides entire JobCards whose projects all fail the predicate, so the user doesn't see empty card headers when the filter is on.
  - Small inline hint when the filter is active but no jobs match, explaining how to clear it.
  - Stateful `title` tooltip on the clickable badge (`Click to show only projects with open PRs` / `Click to clear filter`).

## Follow-ups (separate issues/PRs, not blocking this one)

- **Other 5 header badges** (Failed / Scheduled / Running / Completed / With Issues): the interaction model established here extends trivially to the rest via the same `StatBadge` props. Happy to open follow-up issues/PRs for each once this one lands — keeping them separate makes review easier and lets us iterate on the pattern per-badge if needed.
- **Per-card `hideNoPRs` formula alignment**: the existing per-card `hideNoPRs` filter (around `src/static/index.html:1201`) includes `automerged` in its predicate, while `stats.openPRs` and this PR's new filter do not. The new filter matches `stats.openPRs` exactly for UI consistency. Aligning the per-card filter is out of scope here; happy to open a separate PR if you'd like.

## Pre-existing `just check` failure on `main`

Heads up, unrelated to this PR: `just check` is currently failing on `main` due to pre-existing dead code in `src/controllers/renovatejob_controller_test.go:101,105` (`fakeGitProviderClient` + its `IsFork` method, orphaned by commit `26bc839 refactor: dedicated clients for provider`). This PR doesn't touch any Go code; all 440 unit tests still pass via `just test-unit`. Happy to open a separate cleanup PR if you'd like.

## Test plan

No frontend test suite exists, so validation was manual. I deployed the patched image to my own Kubernetes cluster and walked through a 15-item checklist covering: click toggle + persistence, active state in light/dark mode, keyboard navigation, tooltip content, backward compat on the other badges, per-card filter composition, hide-empty-jobs behavior, "no matches" inline hint, no console errors, clean operator startup logs.

Go unit tests: `just test-unit` → **440 pass, 0 fail, 2 skip**.

## Full disclosure

I don't have any programming skills so I asked Claude Code to implement this and did my best to test it and not create AI slop. I deployed the patched image to my own cluster and verified the fix end-to-end (including the hide-empty-jobs behavior) before opening this PR.

### DIFF
--- a/src/static/components/StatBadge.js
+++ b/src/static/components/StatBadge.js
@@ -5,7 +5,8 @@ export function StatBadge({
   borderClass = "border-gray-200 dark:border-slate-600",
   onClick,
   active = false,
-  activeBorderClass = "ring-2 ring-blue-500 border-blue-500 dark:ring-blue-400 dark:border-blue-400",
+  activeClass = "bg-blue-50 border-blue-500 dark:bg-blue-900/30 dark:border-blue-400",
+  title,
 }) {
   const clickable = typeof onClick === "function";
   const handleKeyDown = clickable
@@ -23,14 +24,16 @@ export function StatBadge({
         onClick,
         onKeyDown: handleKeyDown,
         "aria-pressed": active,
+        title,
       }
     : {};
-  const effectiveBorder = active ? activeBorderClass : borderClass;
+  const bgClass = active ? "" : "bg-white dark:bg-slate-700";
+  const effectiveBorder = active ? activeClass : borderClass;
   const cursorClass = clickable ? "cursor-pointer select-none" : "";
   return (
     <div
       {...interactiveProps}
-      className={`flex items-center justify-between gap-1.5 min-w-[7rem] bg-white dark:bg-slate-700 rounded-lg border ${effectiveBorder} px-3 py-1.5 shadow-sm hover:shadow-md transition-all ${cursorClass}`}
+      className={`flex items-center justify-between gap-1.5 min-w-[7rem] ${bgClass} rounded-lg border ${effectiveBorder} px-3 py-1.5 shadow-sm hover:shadow-md transition-all ${cursorClass}`}
     >
       <span className="text-[0.625rem] text-gray-500 dark:text-slate-400 uppercase tracking-wider whitespace-nowrap">{label}</span>
       <span className={`text-sm font-bold ${valueClass} tabular-nums`}>{value}</span>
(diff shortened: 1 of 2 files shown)


lukashankeln:
Looking pretty good to me 😄 

I especially like the idea itself, we could also use that on the logs-page with the badges we have there 👍 

@beneiltis can you give it a try? My only "fear" is, that it feels strange if a Job-Card just completly disapears.
For me this solution is good though, because the filter is global, so it can be deactivated without a job being present. 

But what do you think?

lukashankeln:
@vtmocanu & @beneiltis we have another possible solution for this issue here: #277 
Can you take a look?

vtmocanu:
@lukashankeln took it for a spin, looks better than mine, have only a small comment, will post it there, also complete implementation, not just PRs badges, I'd say let's evaluate that one and when/if it gets merged I'll close this one.

beneiltis:
Realy honorable vlad! Pleasure to work with you.

vtmocanu:
Closing in favor of https://github.com/mogenius/renovate-operator/pull/277

PR #277, opened 2026-04-09
feat: turn stat badges into filter buttons

### LINKED ISSUE #275: badges as filters - PR badge only PoC

RonaldPhilipsen:
Turn stat badges into filter buttons

image

image

This should make it easier to figure out failures when there's a lot of repositories

### DIFF
--- a/src/static/components/StatBadge.js
+++ b/src/static/components/StatBadge.js
@@ -1,6 +1,40 @@
-export function StatBadge({ label, value, valueClass = "text-gray-900 dark:text-slate-100", borderClass = "border-gray-200 dark:border-slate-600" }) {
+export function StatBadge({
+  label,
+  value,
+  valueClass = "text-gray-900 dark:text-slate-100",
+  borderClass = "border-gray-200 dark:border-slate-600",
+  onClick,
+  selected = false,
+  title,
+}) {
+  const isInteractive = typeof onClick === "function";
+  const className = `flex items-center justify-between gap-1.5 min-w-[7rem] rounded-lg border px-3 py-1.5 shadow-sm transition-all ${borderClass} ${
+    isInteractive
+      ? "bg-white dark:bg-slate-700 hover:shadow-md hover:-translate-y-0.5 cursor-pointer focus:outline-none focus:ring-2 focus:ring-primary/40"
+      : "bg-white dark:bg-slate-700 hover:shadow-md"
+  } ${
+    selected
+      ? "border-primary bg-primary/5 dark:bg-primary/10 ring-2 ring-primary/30"
+      : ""
+  }`;
+
+  if (isInteractive) {
+    return (
+      <button
+        type="button"
+        onClick={onClick}
+        aria-pressed={selected}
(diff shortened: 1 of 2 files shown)


RonaldPhilipsen:
Looks like i reimplemented https://github.com/mogenius/renovate-operator/pull/276 without looking, feel free to close this in favour of that

vtmocanu:
Hey @RonaldPhilipsen, your implementation is better than mine, I really like it, only one small comment/nit:
When I click the PR badge all badges shift a bit to the left, think it would be better if they do not shift on press, what do you think?

Investigated a bit with Claude and it found it to be related to the Clear Filter button and proposed to reserve the space for the button even when hidden (render it always, toggle visibility with opacity-0 pointer-events-none when inactive)

Other than that, congrats, looks good!

lukashankeln:
> Hey @RonaldPhilipsen, your implementation is better than mine, I really like it, only one small comment/nit: When I click the PR badge all badges shift a bit to the left, think it would be better if they do not shift on press, what do you think?
> 
> Investigated a bit with Claude and it found it to be related to the Clear Filter button and proposed to reserve the space for the button even when hidden (render it always, toggle visibility with opacity-0 pointer-events-none when inactive)
> 
> Other than that, congrats, looks good!

Good catch!

Question: Do we need that "Clear Filter" Button in the Header? The Banner that tells you what you filtered (which I really like tbh) also has a clear filter. 
So I think we wouldnt need that one in the Header. wdyt?

vtmocanu:
Yeah, I agree, it is duplicated, also vote to remove the Clear Filter badge, would solve the badges shifting and would remove duplication.

beneiltis:
> Question: Do we need that "Clear Filter" Button in the Header? The Banner that tells you what you filtered (which I really like tbh) also has a clear filter.
> So I think we wouldnt need that one in the Header. wdyt?

Looks good to me but i think the same way like @lukashankeln. Extra button in the header is not necessary. Banner is sufficient. Rest is cool :-)

RonaldPhilipsen:
gotcha, I'll rework this later tonight

RonaldPhilipsen:
> Yeah, I agree, it is duplicated, also vote to remove the Clear Filter badge, would solve the badges shifting and would remove duplication.

Removed the clear filter button, also cleaned up the initialization of `stats` to always start with `openPRs` and `withIssues` 0

RonaldPhilipsen:
closes #275

vtmocanu:
Tested it, looks good to me 👍

lukashankeln (approved):

Will be releasing it with some other stuff in v4.3.0 👍

PR #288, opened 2026-04-14
fix(operator): honor OS theme changes and allow returning to system theme

vtmocanu:
## What

The `system` theme only applied `prefers-color-scheme` on initial page load. Flipping the OS theme with the page open did nothing until reload, and the toggle button switched `light ↔ dark` only, so once clicked there was no way back to `system`.

This PR:
- Adds a `matchMedia` `change` listener that re-applies the theme when the stored value is `system`.
- Makes the header button cycle `system → light → dark → system`. Icon and `aria-label` reflect the current state (sun / moon / monitor).
- Moves `ThemeUtils` and `ThemeToggle` into shared files (`src/static/js/theme-utils.js`, `src/static/components/ThemeToggle.js`) so both `index.html` and `pages/logs.html` use the same code. Claude Code suggested this as best practice after noticing `logs.html` had its own duplicated copy that would otherwise need to be kept in sync.

## Validation

Deployed the built image to my own cluster. OS light↔dark flip propagates live on both `/` and `/logs`, 3-way cycle works, state persists. `just test-unit` passes.

## Full disclosure

I don't have any programming skills so I asked Claude Code to implement this and did my best to test it and not create AI slop. I deployed the patched image to my own cluster and verified both the OS live-update and the 3-way cycle end-to-end before opening this PR.

### DIFF
--- a/src/static/components/ThemeToggle.js
+++ b/src/static/components/ThemeToggle.js
@@ -1,0 +1,53 @@
+export function ThemeToggle() {
+  const [theme, setTheme] = React.useState(() => window.ThemeUtils.getTheme());
+
+  const cycleTheme = () => {
+    const newTheme = window.ThemeUtils.nextTheme(theme);
+    setTheme(newTheme);
+    window.ThemeUtils.setTheme(newTheme);
+    window.ThemeUtils.apply(newTheme);
+  };
+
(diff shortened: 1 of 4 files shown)


lukashankeln (approved):
Once again thank you for your contribution 😄

PR #301, opened 2026-04-19
feat: add valkey configuration to RenovateJobs

RonaldPhilipsen:
This PR Provisions the `RENOVATE_REDIS_URL` on all jobs, reusing the existing valkey configuration.

### DIFF
--- a/src/controllers/renovatejob_controller_test.go
+++ b/src/controllers/renovatejob_controller_test.go
@@ -97,15 +97,6 @@ func (f *fakeDiscovery) WaitForDiscoveryJob(ctx context.Context, job *api.Renova
 	return []string{}, nil
 }
 
-// fakeGitProviderClient implements gitprovider.GitProviderClient for tests.
-type fakeGitProviderClient struct {
-	isForkFn func(ctx context.Context, project string) (bool, error)
-}
-
-func (f *fakeGitProviderClient) IsFork(ctx context.Context, project string) (bool, error) {
-	return f.isForkFn(ctx, project)
-}
-
 type fakeScheduler struct {
 	addedExpr    string
 	addedName    string
--- a/src/internal/renovate/jobDefinitions.go
+++ b/src/internal/renovate/jobDefinitions.go
@@ -6,6 +6,7 @@ import (
 	api "renovate-operator/api/v1alpha1"
 	"renovate-operator/config"
 	crdmanager "renovate-operator/internal/crdManager"
+	"renovate-operator/internal/kvstore"
 	"renovate-operator/internal/utils"
 	"strconv"
 	"strings"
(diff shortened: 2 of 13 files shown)


src/internal/renovate/jobDefinitions.go
```diff
@@ -209,6 +210,22 @@ func getDefaultEnvVars(job *api.RenovateJob) []v1.EnvVar {
...
+			config.GetValue("VALKEY_PORT"),
+			config.GetValue("VALKEY_PASSWORD"),
+		)
+	}
+
+	if redisURL != "" {
+		predefinedEnvVars = append(predefinedEnvVars, v1.EnvVar{
+			Name:  "RENOVATE_REDIS_URL",
+			Value: redisURL,
+		})
```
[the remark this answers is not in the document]
RonaldPhilipsen: This is one to think about, we could create a secret here, but that will require additional permissions on the SA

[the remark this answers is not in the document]
lukashankeln: I think I already added create and update for secrets. (In preparation of the Github App support)

I would also prefer this to be a secret. I think a little Part that does a one-time reconciliation of the secret at startup should be enough.

Also we should try to separate the renovate-cache-data and the session tokens a bit.
I think we could move the renovate-cache into another db within this valkey: https://docs.renovatebot.com/self-hosted-configuration/#redisurl e.g. setting `{{redisUrl}}/1`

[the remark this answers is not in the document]
RonaldPhilipsen: resolved in [56cb8d3](https://github.com/mogenius/renovate-operator/pull/301/commits/56cb8d35916114b9616206645eefbd94caf4a65b)

src/controllers/renovatejob_controller_test.go
RonaldPhilipsen, line 1: Technically out of scope for this PR but this was failing the linting, @lukashankeln lmk if you want me to revert this chainge

lukashankeln (approved):

src/internal/renovate/jobDefinitions.go
```diff
@@ -209,6 +210,20 @@ func getDefaultEnvVars(job *api.RenovateJob) []v1.EnvVar {
 			Value: "debug",
 		})
 	}
+
+	if getValkeyURL() != "" {
```
viceice, line 214: will this override existing redis url env which is set manually?

I think redis prefix should also be set, so multiple jobs don't conflict when they use same redis instance.

https://docs.renovatebot.com/self-hosted-configuration/#redisprefix

RonaldPhilipsen: Ideally you want all jobs to share the same redis cache to get the most use out of it, if they make the same request, you can proxy that answer. could be a nice optional feature though

lukashankeln: User-defined env-vars always overwrite the defaults that the operator sets.

Also we added a helm value to disable the forwarding of valkey

```yaml
config:
  forwardCacheToJobs: true
```

Also i agree, that by default all jobs should use the same cache. But if you want to split them you can set the redis-prefix in the renovatejob itself

PR #306, opened 2026-04-21
feat(operator): issue-303 enabling adding of extraManifests in the operator values

mahmoud-mahdi:
adding support for extraManifests generic mechanism in the Helm chart values.

This would allow users to define arbitrary Kubernetes manifests directly in values.yaml, which are then rendered and deployed as part of the same Helm release.


lukashankeln (approved):

PR #308, opened 2026-04-21
feat: Add ugprade hooks

pierreyves-lebrun:
# Summary
                                                                                                                                               
  - Adds `preUpgrade` and `postUpgrade` hooks to RenovateJobSpec, allowing users to run arbitrary init/cleanup containers before and after each Renovate execution (similarly to the official helm chart [preCommand and postCommand](https://github.com/renovatebot/helm-charts/blob/main/charts/renovate/values.yaml#L63-L72))
  - Introduces a `SubStatus` field on `ProjectStatus` to track which phase a project is currently in (`preUpgrade`, `postUpgrade`, or empty for the main Renovate run)
  - Executor state machine extended: `dispatchScheduled` launches the `preUpgrade` hook instead of the Renovate job when configured, reconcileRunning monitors hook jobs and advances the pipeline — `preUpgrade` → `renovate` → `postUpgrade` → final status
  - Hook failures are configurable: `failOnError`: true (default) marks the project failed; `failOnError`: false logs a warning and proceeds
  - Hook jobs reuse the existing job infrastructure: `TTL cleanup`, `DELETE_SUCCESSFUL_JOBS`, `controller ownership`, `label-based selection`, and pod-level settings (`node selector`, `affinity`, `security contexts`, `image pull secrets`) are all inherited

### DIFF
--- a/charts/renovate-operator/crd/renovate-operator.mogenius.com_renovatejobs.yaml
+++ b/charts/renovate-operator/crd/renovate-operator.mogenius.com_renovatejobs.yaml
@@ -3187,6 +3187,708 @@ spec:
                 description: Maximum number of projects to process in parallel
                 format: int32
                 type: integer
+              postUpgrade:
+                description: Hook to run after renovate execution
+                properties:
+                  args:
+                    description: Arguments for the command
+                    items:
+                      type: string
+                    type: array
+                  command:
+                    description: Command to execute (e.g., ["/bin/sh", "-c"])
+                    items:
+                      type: string
+                    type: array
+                  env:
+                    description: Environment variables for the hook
+                    items:
+                      description: EnvVar represents an environment variable present
+                        in a Container.
+                      properties:
+                        name:
+                          description: |-
+                            Name of the environment variable.
+                            May consist of any printable ASCII characters except '='.
+                          type: string
+                        value:
+                          description: |-
+                            Variable references $(VAR_NAME) are expanded
+                            using the previously defined environment variables in the container and
+                            any service environment variables. If a variable cannot be resolved,
+                            the reference in the input string will be unchanged. Double $$ are reduced
+                            to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e.
+                            "$$(VAR_NAME)" will produce the string literal "$(VAR_NAME)".
+                            Escaped references will never be expanded, regardless of whether the variable
+                            exists or not.
+                            Defaults to "".
+                          type: string
+                        valueFrom:
+                          description: Source for the environment variable's value.
+                            Cannot be used if value is not empty.
+                          properties:
+                            configMapKeyRef:
+                              description: Selects a key of a ConfigMap.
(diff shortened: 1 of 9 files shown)


beneiltis:
hey @pierreyves-lebrun sorry for the delayed response. we were asking ourselves why you need the operator to do this? because in the renovate config you could do something like:
image

If you are still interessted in this: please update your PR and make it more clear for us why this is needed in the operator. otherwise we would like to close this PR.

pierreyves-lebrun:
> hey @pierreyves-lebrun sorry for the delayed response. we were asking ourselves why you need the operator to do this? because in the renovate config you could do something like: image
> 
> If you are still interessted in this: please update your PR and make it more clear for us why this is needed in the operator. otherwise we would like to close this PR.

Thanks for the feedback!

The reason `postUpgradeTasks` doesn't cut it is sort of explained in [the docs](https://docs.renovatebot.com/configuration-options/#postupgradetaskscommands):
`A list of post-upgrade commands that are executed before a commit is made by Renovate.`

That feature and useful and I actually am using it already, but it only solves some specific use cases.

Consider the cases where commands need to be run before and after the commit or PR is raised. 

This is what the official helm chart [preCommand and postCommand](https://github.com/renovatebot/helm-charts/blob/main/charts/renovate/values.yaml#L63-L72) flags solve and and this is what I replicated in this PR.

I am open to amending it if you find any gap. Just for your info, I've been running my fork in prod for the last couple weeks with no issue.

Edit: `postUpgradeTasks` technically runs the specified commands _**while**_ renovate runs, as an optional step among many others. This is either too early or too late for some use cases.

lukashankeln:
But what is the actual use case? As far as I can tell it was originally added to the renovate chart to support GitHub App token renewals? But the Operator already can handle that natively.

Maybe we got usecases here that are better suited for first class operator support?

beneiltis:
Hi @pierreyves-lebrun, thanks again for the effort you put into this — it’s a well-built PR and we appreciate the detailed follow-up.
After discussing it internally, we’ve decided not to merge this and to close the PR. Our reasoning:

- The concrete use case is still unclear to us.The helm chart’spreCommand/postCommandmainly existed to work around things like GitHub App token renewal, which the operator already handles natively. Without a specific scenario that can’t be solved otherwise, we can’t judge whether generic hooks are the right abstraction.

- It’s a significant maintenance commitment.The change extends the CRD API surface and the executor state machine (preUpgrade→renovate→postUpgrade, sub-status tracking, failure semantics). Generic escape hatches like arbitrary hook containers are something we’d have to support indefinitely, and they tend to hide use cases we’d rather support as first-class operator features.

That last point is genuine: if you can share what you’re actually running in your hooks, we’d be happy to discuss it in an issue — chances are it’s something the operator should support natively, in a way that works for everyone.
Since you’re running your fork in production, closing this shouldn’t block you. Thanks again for contributing, and sorry it didn’t work out this time!

PR #313, opened 2026-04-22
fix(crds): add Argo CD hooks

volker-raschek:
The following patch adds Argo CD-specific hooks, as currently there are only hooks for Helm.

This means that during the initial deployment via Argo CD, the CRDs are not present and the sync fails.

Further information can be found in the documentation of the Argo CD hooks: https://argo-cd.readthedocs.io/en/release-2.9/user-guide/resource_hooks/

### DIFF
--- a/charts/renovate-operator/Chart.yaml
+++ b/charts/renovate-operator/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: renovate-operator
 description: Chart to deploy the renovate operator to your cluster
 type: application
-version: 4.8.0
-appVersion: 4.8.0
+version: 4.8.1
+appVersion: 4.8.1
 home: https://mogenius.com
 icon: https://avatars.githubusercontent.com/u/92862356
 sources:
--- a/charts/renovate-operator/templates/crd-hook.yaml
+++ b/charts/renovate-operator/templates/crd-hook.yaml
@@ -6,6 +6,9 @@ metadata:
   name: {{ include "renovate-operator.fullname" . }}-crd-hook
   namespace: {{ .Release.Namespace }}
   annotations:
+    argocd.argoproj.io/hook: PreSync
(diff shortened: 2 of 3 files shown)


lukashankeln:
ArgoCD also supports Helm-Hooks - https://argo-cd.readthedocs.io/en/latest/user-guide/helm/#helm-hooks

The currently selected Helm-Hooks should match to ArgoCDs PreSync Hooks.

What was the exact error you where experiencing?

If you are running with ArgoCD you could also set the following Helm Values to skip the Hook entirely

```yaml
crd:
  mode: template
```
(The default is the hook to enable native Helm users)

volker-raschek:
Hi @lukashankeln, 
that's right, but there is also mentioned a warning:

> [!WARNING]
> Helm hooks + ArgoCD hooks
>
> If you define any Argo CD hooks, all Helm hooks will be ignored.

At least one Argo CD [annotation](https://github.com/mogenius/renovate-operator/pull/313/changes#diff-b53ba302e0066c218a21196b8b2880861ae369bdd0b5a0d9e563aaed7d51a4e9L49) is already present, which, as far as I understand, leads to ignorance of the helm annotations and hooks.

To avoid in future, that someone add a single Argo CD annotation to declare a hook, which breaks the helm hooks, I've added the Argo CD annotations as well to the other resources where helm hooks are defined.

> What was the exact error you where experiencing?

```text
The Kubernetes API could not find renovate-operator.mogenius.com/RenovateJob for requested resource renovate-operator/github. Make sure the "RenovateJob" CRD is installed on the destination cluster.
```

To my surprise, I didn't see any kubectl pod during the deployment in Argo CD. Normally, they pop up briefly. But here, it seems kubectl pod wasn't executed at all, which is why the CRD is missing.

If I deploy the operator via Helm, the CRDs deploy successfully :smile:

PR #318, opened 2026-04-29
fix: add-cancel-job-button-and-represnent-prioritization-in-ui'

beneiltis:
1. you can now cancel running jobs. 
Screenshot 2026-04-29 at 11 57 03

2. when you now prioritize a task it will also be shown in the UI. currently it is prioritized but not represented in the UI.


src/internal/crdManager/renovateJobManager.go
```diff
@@ -441,6 +444,49 @@ func (r *renovateJobManager) IsWebhookSignatureValid(ctx context.Context, job Re
...
+		JobType:   ExecutorJobType,
+		Namespace: job.Namespace,
+	})
+	if err == nil && executorJob != nil {
+		if delErr := DeleteJob(ctx, r.client, executorJob); delErr != nil {
+			return fmt.Errorf("failed to delete executor job: %w", delErr)
+		}
+	}
+
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
lukashankeln: For the actual status update I think we could use `UpdateProjectStatus`

beneiltis: thx, absolutly right :-)

lukashankeln (approved):

PR #347, opened 2026-05-20
fix(ui): remove horizontal scrolling on dashboard

derdiggn:
The dashboard scrolled horizontally at a bunch of viewport widths, for a few unrelated reasons.

On narrow mobile, the "Renovate: The Kubernetes-Native Way" tagline had `whitespace-nowrap`, so the header alone overflowed at around 340px. Dropping the class lets it wrap onto two lines, which is what you'd expect on a phone.

Between 768px and 1024px the project list was already trying to render as the seven-column desktop table, and there just isn't room for that in this range. The table-vs-card swap is now at the `lg:` breakpoint (1024px) instead of `md:`. Below that, you get the card view mobile already used; above, the table.

At the `lg:` boundary the table only barely fits, so cell padding now drops to `px-3` from `lg:` and goes back to `px-6` from `xl:`. That freed up about 170 pixels of column padding at the tight end, enough to clear the remaining overflow.

One more, smaller thing: the Actions column header had a full "Trigger All" button packed in next to its label, which widened the column on every row. The same button already lived up in the job card header behind `md:hidden`. I removed the `md:hidden`, deleted the duplicate inside the `<th>`, and now "Trigger All" sits beside "Run Discovery" at every width.

Checked across the breakpoint boundaries and a few widths in between:

| Viewport | Document overflow |
| --- | --- |
| 340px | 0 |
| 768px | 0 |
| 1023px | 0 |
| 1024px (table appears) | 0 |
| 1280px (full padding) | 0 |
| 1920px | 0 |

Files: `src/static/index.html`, `src/static/components/SiteHeader.js`.


derdiggn:
after the changes: 

after-340
after
after-790
after-1252

lukashankeln:

Looking good, but at least for me this results in a strange wrapping in the header. Can we extend the width of that part a little bit?
image

derdiggn:
@lukashankeln , Was thinking of tightening up the header a bit more: moving the version number to its own line, and adjusting the light/dark mode toggle position on mobile so we make better use of the space.

lukashankeln:
> @lukashankeln , Was thinking of tightening up the header a bit more: moving the version number to its own line, and adjusting the light/dark mode toggle position on mobile so we make better use of the space.

Sounds good 👍

derdiggn:
Your comment sent me down a rabbit hole. The tagline wrap was a symptom of a header that was trying to do too much in one flex row, so I redid it as two strips:

- Brand strip with the logo, tagline + version, theme toggle (and auth when it's enabled). Below `md:` this stacks as two rows: logo + theme on top, tagline + version below. From `md:` upward everything collapses inline.
- Stats strip with the 8 filter badges, in its own band separated by a subtle border.

While I was in there I also picked off the smaller things:

- `Kubernetes-Native` used to break at its internal hyphen. `whitespace-nowrap` on the inner span keeps it together now.
- The version badge crowded the tagline at certain widths. It drops to its own line when the row gets squeezed.
- The theme toggle was floating opposite the tagline on mobile. It sits next to the logo on row 1 instead, including at 640 px.
- The 8 stat badges always wrapped to 5+3 at desktop widths because of a forced `lg:` line break. Dropped it. The container is `grid grid-cols-2` below `sm:` (badges stretch to fill the column) and `flex flex-wrap` from `sm:` upward. So mobile gets 4 rows of 2, and from ~960 px on all 8 sit on a single row.

Before / after at six widths:

| Width | Before | After |
| --- | --- | --- |
| 340 px  |   |   |
| 640 px  |   |   |
| 768 px  |   |   |
| 1024 px |  |  |
| 1280 px |  |  |
| 1920 px |  |  |

No horizontal page scroll at any of those widths. The screenshots commit (`f246170`) can be reverted or squashed away before merge if you'd rather not carry the PNGs in the repo.

lukashankeln (approved):

beneiltis (approved):

PR #351, opened 2026-05-22
fix(deployment): support custom valkey auth secret

volker-raschek:
The upstream Helm chart "valkey" supports an alternative custom secret that contains the users' passwords via `auth.usersExistingSecret`. Additionally provides the helm chart a configuration property to define the key inside the secret via `auth.aclUsers.default.passwordKey`.

https://github.com/valkey-io/valkey-helm/blob/39a708f93a145b5cd254e9f36ffa735d929a08d0/valkey/values.yaml#L183

```yaml
auth:
  # Use an existing secret for user passwords. Key defaults to username.
  usersExistingSecret: ""

  # Map of users to create with ACL permissions.
  # If usersExistingSecret is set, passwords from the secret take priority over inline passwords.
  # NOTE: If using aclUsers, the 'default' user must be included here.
  aclUsers: {}
  # Example:
  # default:
  #   permissions: "~* &* +@all"
  #   password: "secretpass"        # Inline password (fallback if usersExistingSecret not set)
  #   passwordKey: "admin-pwd"      # Key in usersExistingSecret (defaults to username)

```

The deployment of the renovate-operator fails, if `valkey.auth.enabled` is set to `true`, because valkey expect that either `auth.aclUsers.default.password` is set or an existing secret is defined via `auth.usersExistingSecret`.

If `auth.usersExistingSecret` is used, the deployment of the renovate-operator fails, because the hard coded secret `{{ .ReleaseName }}-valkey-auth` can not be found.

The upstream properties `auth.usersExistingSecret` and `auth.aclUsers.default.passwordKey` will now be respected.

I've added an addtional step into the GitHub Action workflow `helm-lint` to install helm unittest to execute the implemented unittests defined in `charts/renovate-operator/unittests`. These tests only verify whether my change is templated as intended. They do not take anything else into account. However, this change lays the groundwork for future unit tests.

Since the OSS version of VSCode does not load the Helm unittest YAML schema, I created a settings.json file that adds the YAML schema to the project and will help developers write unit tests in the future.

### DIFF
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
-        
+
       - name: Install Helm
         uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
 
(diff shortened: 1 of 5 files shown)


lukashankeln (approved):

PR #377, opened 2026-06-03
fix: delete successful discovery jobs when DELETE_SUCCESSFUL_JOBS=true

### LINKED ISSUE #327: `DELETE_SUCCESSFUL_JOBS` doesn't seem to do anything
**Describe the bug**
setting `DELETE_SUCCESSFUL_JOBS=true` via chart `config.deleteSuccessfulJobs=true` doesn't remove the jobs

**To Reproduce**
1. set `config.deleteSuccessfulJobs=true` in chart
2. wait for jobs succeed
3. see jobs are not removed

**Expected behavior**
That settings seems to remove successful jobs, so it should do that.

**Screenshots & Logs**

 - Kubernetes Version v1.34.7+k3s1
 - Version 4.7.0

**Additional context**
using valkey for distributed mode (session, logs, renovate cache)

sridhar-3009:
## Problem

Closes #327.

`DELETE_SUCCESSFUL_JOBS=true` (set via `config.deleteSuccessfulJobs=true` in the Helm chart) only deleted **executor** (per-project) K8s `batch/v1/Job` resources through `reconcileRunning` in `executor.go`. **Discovery** K8s Jobs were never deleted, so they accumulated indefinitely even when the setting was explicitly enabled.

## Root cause

`executor.go` line 240 correctly deletes executor jobs:
```go
if newStatus == api.JobStatusCompleted && config.GetValue("DELETE_SUCCESSFUL_JOBS") == "true" && k8sJob != nil {
    crdManager.DeleteJob(ctx, e.client, k8sJob)
}
```

But `discoveryAgent.go`'s `WaitForDiscoveryJob` fetches the completed discovery K8s Job (to read project logs) and then returns — never deleting it.

## Fix

After extracting discovered projects from the completed discovery job's logs, call `crdManager.DeleteJob` if `DELETE_SUCCESSFUL_JOBS=true`:

```go
if config.GetValue("DELETE_SUCCESSFUL_JOBS") == "true" {
    if err := crdManager.DeleteJob(ctx, e.client, existingDiscoveryJob); err != nil {
        log.FromContext(ctx).Error(err, "failed to delete successful discovery job", ...)
    }
}
```

Errors during deletion are logged but do not abort the discovery flow — projects have already been extracted at that point.

### DIFF
--- a/src/internal/renovate/discoveryAgent.go
+++ b/src/internal/renovate/discoveryAgent.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	api "renovate-operator/api/v1alpha1"
 	crdManager "renovate-operator/internal/crdManager"
+	"renovate-operator/internal/config"
 	"renovate-operator/internal/telemetry"
 	"renovate-operator/internal/utils"
 	"sync"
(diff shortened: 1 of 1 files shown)


lukashankeln (approved):
Thank you for your contribution

PR #379, opened 2026-06-03
fix: delete successful discovery jobs when DELETE_SUCCESSFUL_JOBS=true

sridhar-3009:
## Summary
- Adds handling for `DELETE_SUCCESSFUL_JOBS` env var in discovery job cleanup
- When set to `true`, successfully completed discovery jobs are deleted after processing
- Fixes config import path (`renovate-operator/config` not `renovate-operator/internal/config`)

## Test plan
- [ ] Set `DELETE_SUCCESSFUL_JOBS=true` and verify completed discovery jobs are cleaned up
- [ ] Set `DELETE_SUCCESSFUL_JOBS=false` (or unset) and verify jobs are retained


lukashankeln:
Didnt we already fix this?

PR #388, opened 2026-06-12
feat: skip pending-deletion repos during discovery

tibuntu:
## Problem

When using autodiscovery against a self-hosted GitLab, Renovate jobs are created even for
repositories GitLab has marked for **delayed deletion**. While in that state a project is still
discoverable, but running Renovate against a repo that's on its way out is wasteful and noisy.

## Solution

A single opt-in `RenovateJobSpec` field, `skipPendingDeletion`, mirroring the existing `skipForks`
end to end. When set, repositories marked for deletion are excluded after autodiscovery, before
execution jobs are created. Defaults to `false`, so existing behavior is unchanged.

## Design

Fork and pending-deletion detection are unified behind a single interface method,
`GetRepositoryInfo(ctx, project) (RepositoryInfo, error)` returning `{Fork, PendingDeletion}`,
and a single `FilterProjects` pass (in `projectFilter.go`, replacing the old `forkFilter.go`).
This means enabling both `skipForks` and `skipPendingDeletion` fetches each repository's metadata
in **one** platform API call rather than two. The pass is concurrent (max 10) and **fail-open** on
per-repo API errors, and is wired into `ReconcileProjects` after discovery.

| Platform    | Fork detection            | Pending-deletion detection                              |
|-------------|---------------------------|---------------------------------------------------------|
| `gitlab`    | `forked_from_project`     | `marked_for_deletion_at` / `marked_for_deletion_on`     |
| `github`    | `fork`                    | not supported (deletes are immediate) — `false`         |
| `gitea`     | `fork`                    | not supported — `false`                                 |
| `forgejo`   | `fork`                    | not supported — `false`                                 |
| `bitbucket` | `parent`                  | not supported — `false`                                 |

All fields come from the provider's existing single repo-metadata endpoint (e.g. GitLab
`GET /projects/{path}`). Only GitLab exposes a pending-deletion state.

> Note: archived repositories are intentionally **not** handled here — they are already skipped
> during onboarding.

## Tests

- `projectFilter_test.go` — table-driven: each criterion in isolation, both combined, both-off no-op, and per-repo API error → kept (fail-open).
- `gitlabProvider/gitlab_test.go` — httptest covering fork, both `marked_for_deletion_*` variants, empty-field-not-pending, and the combined case.
- Existing mocks updated to the unified interface.

## Docs

`docs/autodiscovery.md` gains an "Excluding Pending-Deletion Repositories" section; CRD manifest updated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tibuntu:
Addressed the double-API-call feedback: fork and pending-deletion detection are now unified behind a single `GetRepositoryInfo` method and one `FilterProjects` pass, so enabling both `skipForks` and `skipPendingDeletion` fetches each repo's metadata in a single `GET /projects/{path}` call instead of two. `forkFilter.go` was replaced by `projectFilter.go` accordingly.

lukashankeln (approved):

PR #404, opened 2026-06-16
feat: add support for running helm unittests through just

tobiaskohlbau:
There is no way to run the helm unittests through just and the default helm unittests does not work either. In order to imrove DX add a just target for convenience.


wittdennis (approved):

lukashankeln:

PR #433, opened 2026-06-21
fix(ui): set Cache-Control headers on static assets

pat-s:
## Summary

The UI's static files are served by a plain `http.FileServer` (`src/ui/ui.go`) with no `Cache-Control` or `ETag` — only `Last-Modified`. Browsers then apply heuristic caching to `index.html` and the scripts, so after a deploy a browser can keep running a **stale `index.html` against changed assets**. In practice this surfaces as confusing console errors from a mismatched cached page, e.g.:

```
Uncaught SyntaxError: export declarations may only appear at top level of a module  (babel.min.js)
Uncaught ReferenceError: StatBadge is not defined
```

even though the deployed files are correct.

## Change

Tiered cache policy in `registerUiRoutes`:

- **`no-cache`** (revalidate every load, cheap 304 via `Last-Modified`) for app-authored files that change each deploy: `index.html`, `/logs`, `/components/*.js`, `/js/babel-config.js`, `/js/theme-utils.js`, `/css/*`.
- **`public, max-age=86400`** for the large version-pinned vendored bundles `/js/babel.min.js`, `/js/react-bundle.esm.js`, `/js/tailwind.min.js` — they only change when their pin is bumped (which ships a new image), so there's no need to revalidate ~3.4MB on every load. The 1-day cap bounds staleness after a rare pin bump.

A follow-up could content-hash the vendored bundles to cache them `immutable`, but they aren't hashed today.

### DIFF
--- a/src/ui/ui.go
+++ b/src/ui/ui.go
@@ -14,27 +14,15 @@ func (s *Server) registerUiRoutes(router *mux.Router) {
 	router.PathPrefix("/").Handler(cacheControl(http.FileServer(http.Dir("./static/"))))
 }
 
-// vendoredAssets are version-pinned third-party bundles served under stable
-// filenames. They change only when their pin is bumped (which ships a new
-// image), so they can be cached. Everything else the UI serves is app-authored
-// and changes every deploy, so it must be revalidated to avoid a browser
-// running a stale index.html against changed scripts.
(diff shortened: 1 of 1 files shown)


lukashankeln (approved):

PR #434, opened 2026-06-21
fix(forgejo): treat 404 as success when deleting a webhook

pat-s:
## Summary

`ForgejoClient.DeleteRepoWebhook` returned an error for any non-2xx status, including `404`. But a webhook that no longer exists is the desired end state of a delete, so a 404 should be treated as success.

In the webhook sync loop, this caused the removal step to error and retry noisily on hooks that were already gone (e.g. deleted manually or by a previous run):

```
failed to remove webhook ... unexpected status 404: {"message":"The target couldn't be found."}
```

## Change

`DeleteRepoWebhook` now returns `nil` on `404`. Added `TestDeleteRepoWebhook_NotFoundIsSuccess`.

### DIFF
--- a/src/gitProviderClients/forgejoProvider/forgejo.go
+++ b/src/gitProviderClients/forgejoProvider/forgejo.go
@@ -172,6 +172,10 @@ func (c *ForgejoClient) DeleteRepoWebhook(ctx context.Context, owner, repo strin
 	}
 	defer func() { _ = resp.Body.Close() }()
 
+	// A missing webhook is the desired end state, so treat 404 as success.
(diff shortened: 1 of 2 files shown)


pat-s:
Addressed the review:

- **forgejo.go:** drain the response body on the 404 path (`io.Copy(io.Discard, resp.Body)`) before returning, so the connection can be reused — Forgejo includes a JSON body on 404.
- **forgejo_test.go:** the not-found test now asserts the request method is `DELETE`, matching `TestDeleteRepoWebhook`.

lukashankeln (approved):
Thx for the contribution

PR #435, opened 2026-06-21
fix(webhook-sync): sync webhooks for autodiscovered repos without a topic

pat-s:
## Summary

`WebhookSyncer.RunOnce` discovered repositories **only** by topic, and `webhookSyncManager.RunSync` called `RunOnce(ctx)` with no repos. So a `RenovateJob` that autodiscovers repositories **without** a configured topic registered no webhooks at all — the sync logged and bailed:

```
no projects provided and no topic configured, skipping webhook sync
```

## Change

- `RunOnce(ctx, projects ...string)` — when a project list is provided, those repos are used directly (admin assumed; repos where admin is missing are detected via the webhook API's 403 and skipped). Falls back to topic search when no projects are passed, and skips cleanly when there's neither.
- `RunSync` loads the `RenovateJob` and passes its discovered `Status.Projects`, so topic-less autodiscover setups sync webhooks.
- Extracted `snapshotManagedRepos()` helper (used by the new early-return and the final return).
- Tests: `TestSyncUsesProvidedProjectsWithoutTopic`, `TestSyncSkipsWhenNoProjectsAndNoTopic`.

### DIFF
--- a/src/internal/webhookSync/manager.go
+++ b/src/internal/webhookSync/manager.go
@@ -144,7 +144,20 @@ func (m *webhookSyncManager) RunSync(ctx context.Context, logger logr.Logger, jo
 	if !ok {
 		return
 	}
-	state, err := entry.syncer.RunOnce(ctx)
+	// Pass the discovered project list so webhook sync works for jobs that
+	// autodiscover repositories without a topic. RunOnce falls back to topic
+	// search when no projects are supplied.
+	var projects []string
+	if rj, err := m.jobManager.GetRenovateJob(ctx, jobId.Name, jobId.Namespace); err != nil {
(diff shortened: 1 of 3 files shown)


pat-s:
Addressed the review (pushed f9df802):

- **sync.go (403 handling):** the provided-projects branch now genuinely skips repos that return `403` from the webhook API (logged at info, not error) — so an autodiscovered list that includes non-admin repos doesn't flood error logs. The comment now matches the behavior, and `repos` is preallocated to `len(projects)`.
- **manager.go (log wording):** reworded to "RunOnce will use topic search if configured, otherwise skip" so it's accurate for topic-less jobs.
- Added `TestSyncSkipsProvidedProjectWhenWebhookAPIForbids` covering the 403 skip.

lukashankeln (approved):
Thx for the fix

PR #446, opened 2026-06-26
feat(operator): expose pending-approval duration as a metric

tibuntu:
## Problem

`renovate_operator_approvals_needed` only reports the current count, with no notion of *how long* approvals have been pending. Because the gauge carries `pod`/`instance` labels and the operator restarts regularly, no single series is long-lived, so duration-based alerting via `min_over_time`/`for` can't reliably express "pending for N days".

## Change

- Persist `ApprovalsNeededSince` in the project status. It is set when approvals first appear, preserved while they remain, and cleared once none are pending. Living in the CRD status, it survives operator restarts.
- Expose it as a gauge `renovate_operator_approvals_needed_since_seconds` (Unix timestamp), emitted after a successful status update and re-hydrated from the durable status on reconcile so it is restored after a restart.

This lets consumers alert on duration directly, e.g.:

```promql
time() - max by (project) (renovate_operator_approvals_needed_since_seconds) > 518400
```

## Notes

- New CRD field is optional; existing resources are unaffected.
- Unit tests cover the set/preserve/clear transitions.

### DIFF
--- a/src/controllers/renovatejob_controller.go
+++ b/src/controllers/renovatejob_controller.go
@@ -7,6 +7,7 @@ import (
 	"renovate-operator/internal/renovate"
 	"renovate-operator/internal/telemetry"
 	"renovate-operator/internal/types"
+	"renovate-operator/metricStore"
 	"renovate-operator/scheduler"
 	"strings"
 	"time"
@@ -64,6 +65,7 @@ func (r *RenovateJobReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 		// renovatejob object read without problem -> create the schedule
 		r.ensureWebhookCleanupFinalizer(ctx, logger, renovateJob)
+		metricStore.RehydrateMetrics(renovateJob.Namespace, renovateJob.Name, renovateJob.Status.Projects)
 		r.resetOrphanedRunning(ctx, renovateJob)
 		createScheduler(logger, renovateJob, r)
 		if err := r.GithubApp.EnsureToken(ctx, renovateJob); err != nil {
--- a/charts/renovate-operator/crd/renovate-operator.mogenius.com_renovatejobs.yaml
+++ b/charts/renovate-operator/crd/renovate-operator.mogenius.com_renovatejobs.yaml
@@ -4287,6 +4287,11 @@ spec:
                 items:
                   description: Status of a single project within a RenovateJob
                   properties:
+                    approvalsNeededSince:
+                      description: ApprovalsNeededSince is when this project's current
+                        streak of pending approvals began, cleared once none are pending.
+                      format: date-time
+                      type: string
                     duration:
                       type: string
                     lastRun:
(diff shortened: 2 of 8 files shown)


src/controllers/renovatejob_controller.go
```diff
@@ -57,6 +58,7 @@ func (r *RenovateJobReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	if err == nil {
 		// renovatejob object read without problem -> create the schedule
+		metricStore.RehydrateApprovalMetrics(renovateJob.Namespace, renovateJob.Name, renovateJob.Status.Projects)
```
lukashankeln: I think we should go with `RehydrateMetrics` here, we might also need it in the future for other onces.

I think we could already also rehydrate `renovate_operator_run_failed` from the crd

src/internal/crdManager/renovateJobManager.go
```diff
@@ -245,18 +250,35 @@ func (r *renovateJobManager) UpdateProjectStatusBatched(ctx context.Context, fn
...
+		}
+		for _, i := range updatedIdx {
+			syncApprovalsNeededSinceMetric(renovateJob.Namespace, renovateJob.Name, renovateJob.Status.Projects[i])
+		}
+		return nil
 	})
 }
 
+// syncApprovalsNeededSinceMetric mirrors a project's durable ApprovalsNeededSince into the gauge.
+func syncApprovalsNeededSinceMetric(namespace, job string, p api.ProjectStatus) {
```
lukashankeln: Is it really needed to move the metrics into the manager here?

For me this feels like the wrong place?

I think we should do this where we already are setting the other metrics: https://github.com/mogenius/renovate-operator/blob/main/src/internal/renovate/executor.go#L232

src/controllers/renovatejob_controller.go
tibuntu: Done. Renamed to RehydrateMetrics and it now also rehydrates renovate_operator_run_failed from the persisted status. Left dependency_issues out for now but it's an easy follow-up via the same path.

src/internal/crdManager/renovateJobManager.go
tibuntu: Agreed, moved it out of the manager. The metric is now set in executor.go next to the other run metrics. The timestamp decision lives in a small pure helper (utils.NextApprovalsNeededSince) that the executor calls; the manager just persists the value via the status update.

tibuntu:
ping @lukashankeln @wittdennis

wittdennis:
With version 5.4.0 `renovate_operator_approvals_needed`  is being re-hydrated and never resets to 0.

This makes the `renovate_operator_approvals_needed_since_seconds` redundant as you can now just create a duration based query over the `renovate_operator_approvals_needed` metric.

Regardless thanks for the contribution even tho we decided to skip this one!

PR #454, opened 2026-06-30
feat(webhook): support Standard Webhooks signature authentication

pdecat:
## What

Adds a third webhook authentication scheme: [Standard Webhooks](https://www.standardwebhooks.com/) signatures, alongside the existing plain-token (`Authorization` / `X-Gitlab-Token`) and GitHub/Forgejo HMAC (`X-Hub-Signature-256` / `X-Forgejo-Signature`) paths.

Motivating use case: GitLab v19 added [signing tokens](https://docs.gitlab.com/user/project/integrations/webhooks/#signing-tokens) — GitLab's recommended alternative to the legacy plaintext secret token — are an implementation of Standard Webhooks. The scheme itself is vendor-neutral, so verification here is done generically and works for any compliant sender, not just GitLab.

## How

- New `IsWebhookStandardSignatureValid` on `RenovateJobManager`: for each configured secret, strip the `whsec_` prefix and base64-decode it into the HMAC key, recompute `HMAC-SHA256` over `"{webhook-id}.{webhook-timestamp}.{body}"`, and constant-time compare against each `v1,<base64>` entry of the `Webhook-Signature` header.
- Rejects deliveries whose `Webhook-Timestamp` is more than 5 minutes from the current time (replay window).
- `buildAuthCheckerFromRequest` routes requests carrying a `Webhook-Signature` header to the new validator, preferring the signature over a plain token when both are present.
- Reuses the existing `authentication.secretRef` secret (comma-separated values), so a signing key and a legacy token can coexist in one secret.

No changes to the provider abstraction or CRD types — this slots into the existing credential-validator interface.

## Tests

- Unit tests: the canonical Standard Webhooks known-answer vector, key-decode edge cases (`whsec_` prefix / bare / invalid), the timestamp window, and header routing/precedence.
- An over-the-wire integration test (build-tagged `integration`, so it stays out of the default unit run) that boots the real webhook server + real manager over a fake k8s client and drives it with an independent signer: valid → 202 + project scheduled, tampered body → 401, wrong key → 401, stale/future timestamp → 401, missing credentials → 401, legacy `X-Gitlab-Token` → 202, auth-disabled job → 202.

`just build`, `just test-unit` (570 pass), and `just test-integration` (8/8) are green.

### DIFF
--- a/src/internal/crdManager/renovateJobManager.go
+++ b/src/internal/crdManager/renovateJobManager.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"crypto/hmac"
 	"crypto/sha256"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
 	"slices"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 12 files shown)


src/internal/crdManager/renovateJobManager.go
```diff
@@ -445,6 +451,56 @@ func (r *renovateJobManager) IsWebhookSignatureValid(ctx context.Context, job Re
...
 }
 
+// IsWebhookStandardSignatureValid validates a Standard Webhooks signature against the keys configured for
+// the RenovateJob. Standard Webhooks (https://www.standardwebhooks.com/) is a vendor-neutral webhook
+// signing scheme implemented by multiple providers — GitLab "signing tokens" among them — so this path is
+// not GitLab-specific and works for any compliant sender. The signed content is "{msgID}.{timestamp}.{body}",
+// keyed by the HMAC-SHA256 key decoded from each configured secret. The signature header is a space-separated
+// list of "v1,<base64>" entries; a match against any entry for any configured key authenticates the request.
+// The timestamp must be within standardWebhookTimestampTolerance of now to reject replayed requests.
+func (r *renovateJobManager) IsWebhookStandardSignatureValid(ctx context.Context, job RenovateJobIdentifier, msgID, timestamp, signature string, body []byte) (bool, error) {
```
pdecat, line 461: Code from this new `IsWebhookStandardSignatureValid` function could be shared with `IsWebhookSignatureValid` to avoid duplication.

An alternative may be to make `IsWebhookSignatureValid` support both types of signatures.

lukashankeln (approved):

PR #472, opened 2026-07-08
feat(ui): make the OIDC groups claim name configurable

pdecat:
The OIDC groups claim is hardcoded to `groups` via Go struct tags on both the ID
token and the userinfo claim sets (`src/ui/oidc.go`). Providers that emit group
memberships under a namespaced claim (e.g. `https://example.com/groups`, common
with Auth0 and some enterprise IdPs) therefore return no groups, so the existing
group filtering (`OIDC_ALLOWED_GROUP_PREFIX` / `OIDC_ALLOWED_GROUP_PATTERN`)
can't be used with them.

- Add `OIDC_GROUPS_CLAIM` (Helm: `auth.oidc.groupsClaim`), defaulting to `groups`
  so existing deployments are unchanged.
- Read the groups claim dynamically by the configured name from both the ID token
  and the userinfo endpoint, tolerating either a JSON array of strings or a single
  string.
- Add a unit test for the claim parser; wire the value through `main.go` and the
  Helm chart (`deployment.yaml` + `values.yaml`).

### DIFF
--- a/charts/renovate-operator/templates/deployment.yaml
+++ b/charts/renovate-operator/templates/deployment.yaml
@@ -171,6 +171,10 @@ spec:
             - name: OIDC_FETCH_USERINFO_GROUPS
               value: "true"
             {{- end }}
+            {{- if .Values.auth.oidc.groupsClaim }}
+            - name: OIDC_GROUPS_CLAIM
+              value: {{ .Values.auth.oidc.groupsClaim | quote }}
+            {{- end }}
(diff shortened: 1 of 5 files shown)


lukashankeln:
@pdecat this looks to me like a good change, is there a reason that the PR is still a Draft?

pdecat:
I was just waiting for a colleague of mine to review it

lukashankeln (approved):

PR #473, opened 2026-07-08
feat(operator): serve UI, API and auth under a configurable sub-path

splattner:
## Summary

Adds support for serving the operator's web UI, API, auth, and health endpoints under a configurable sub-path (e.g. `https://host/renovate`) instead of only the domain root. This lets the operator be co-hosted with other applications on the same hostname behind a single Ingress/Gateway.

Behaviour is fully backward compatible: when the new setting is empty (the default), everything is served from `/` exactly as before.

## Motivation

Previously the UI assumed it was mounted at the domain root. The frontend emitted absolute URLs (`/api/v1/...`, `/js/...`, `/auth/login`), so an Ingress `rewrite-target` alone couldn't make it work — the browser would still request assets and APIs from the host root. This makes it impossible to place the operator under a path prefix next to another app.

## Changes

**Configuration**
- New optional `BASE_PATH` env var (default `""` = root). Normalized to have a leading slash and no trailing slash.

**Backend (`ui` package)**
- All UI, API, auth, and health routes are mounted under a `PathPrefix(basePath)` subrouter; `/` redirects to the base path.
- index.html / logs.html are served through a handler that injects `<base href>` and `window.__BASE_PATH__`.
- Static file serving and vendored-asset cache matching strip the prefix.
- Auth redirects are prefixed via `withBase()`; session/state/PKCE cookies are scoped via `cookiePath()`; the auth middleware strips the prefix before route matching.

**Frontend**
- All runtime URLs (API calls, auth redirects, logs links, assets) are built from the injected `BASE`; head asset references made relative so they resolve against `<base href>`.

**Helm chart**
- New top-level `basePath` value, wired into the `BASE_PATH` env, the Ingress/HTTPRoute path, and the auto-detected OAuth/OIDC redirect URLs.

**Tests**
- Go unit tests for the base-path helpers (`normalizeBasePath`, `withBase`, `cookiePath`).
- Helm unittests asserting the deployment `BASE_PATH` env, Ingress path, and redirect URL rendering.

## Testing

- `go build Other.` and `go test Other.` pass.
- `helm unittest` passes (30 tests, incl. new base-path suites).
- Manually verified against a live cluster with `BASE_PATH=/renovate`:
  - `GET /` → 302 → `/renovate/`
  - `/renovate/` serves index with `<base href="/renovate/">` + `window.__BASE_PATH__`
  - `/renovate/api/v1/version` → 200; `/api/v1/version` at root → 404
  - `/renovate/js/tailwind.min.js` and `/renovate/health` → 200

## Usage

Set `basePath: /renovate` in Helm values.

> ⚠️ **Note:** OAuth/OIDC redirect URLs registered with your identity provider must include the sub-path (e.g. `https://host/renovate/auth/callback`). The chart appends it automatically to the auto-detected value.


wittdennis (approved):
Thanks for the contribution!

Issue #486, opened 2026-07-09
UI renovatejobs endpoint is slow with many RenovateJobs when discovery jobs are missing


benjaminknauer:
**Describe the bug**
  The UI dashboard can take a very long time to load when many `RenovateJob` resources exist and their discovery Kubernetes Jobs are no longer present.

  In an environment with ~50 RenovateJobs using autodiscovery, `/api/v1/renovatejobs` takes around 40-60 seconds. The response is ~800 KB, but the main latency comes from discovery status lookup: each missing
  discovery Job causes `GetDiscoveryJobStatus` to sleep/retry before returning.

  **To Reproduce**
  1. Create multiple `RenovateJob` resources, e.g. 50.
  2. Let autodiscovery populate `status.projects`.
  3. Ensure the corresponding discovery Kubernetes Jobs are absent, e.g. deleted by TTL, cleanup, or `DELETE_SUCCESSFUL_JOBS=true`.
  4. Open the UI dashboard or call `/api/v1/renovatejobs`.
  5. Observe that the request duration grows roughly by one second per RenovateJob without a discovery Job.

  **Expected behavior**
  The dashboard should return quickly. A missing discovery Job is a normal idle/scheduled state and should not cause per-job polling or sleeps during a read-only status request.

  **Screenshots & Logs**
  A regression test reproduces the behavior with 5 RenovateJobs and no discovery Jobs:

  ```text
  --- FAIL: TestGetRenovateJobs_ManyMissingDiscoveryJobsReturnsQuickly (5.07s)
      renovateController_test.go:876: getRenovateJobs took 5.025843419s for 5 jobs without discovery jobs
```

**Information:**

  - Version: 4.14.1

**Additional context**
  status.projects can exist independently from Kubernetes discovery Jobs. Finished discovery Jobs may be deleted by TTL, explicit cleanup, or old-generation cleanup, so NotFound should be handled as a normal
  state for dashboard status reads.

PR #487, opened 2026-07-09
fix(ui): avoid slow renovatejobs loading when discovery jobs are missing

benjaminknauer:
## What changed

  Draft PR for https://github.com/mogenius/renovate-operator/issues/486.

  This adds a regression test for the slow `/api/v1/renovatejobs` path and removes the retry/sleep behavior when a discovery Kubernetes Job is missing.

  The current code calls `GetDiscoveryJobStatus` once per RenovateJob while building the UI response. If the discovery Job no longer exists, that function sleeps/retries before returning. With many
  RenovateJobs this adds up quickly. In our case, ~50 RenovateJobs made the endpoint take around 40-60 seconds.

  ## Why this is a draft

  I am not fully sure what exact race the original sleep workaround was meant to address. From the history it looks like it was added to wait for a discovery Job to appear, probably around manual discovery
  triggering or Kubernetes cache visibility.

  For the dashboard read path, though, a missing discovery Job is a normal state:
  - successful discovery Jobs may be deleted
  - TTL cleanup may remove them
  - old generations may be cleaned up
  - `status.projects` can still be populated even when the discovery Job is gone

  So this PR is mainly meant to show the problem with a focused failing test and propose the smallest fix.

  ## Test added

  The new test covers the real slow path:

  ```bash
  go test ./ui -run TestGetRenovateJobs_ManyMissingDiscoveryJobsReturnsQuickly -count=1
```
  Without the fix, it fails like this:
```
  --- FAIL: TestGetRenovateJobs_ManyMissingDiscoveryJobsReturnsQuickly (5.07s)
      renovateController_test.go:876: getRenovateJobs took 5.025843419s for 5 jobs without discovery jobs, expected it to return without per-job polling
```
  That reproduces the same multiplier we see in larger clusters: roughly one second per RenovateJob without a discovery Job.

  ## Proposed fix

  When GetDiscoveryJobStatus cannot find a discovery Job, it now returns scheduled immediately instead of sleeping/retrying.

  This keeps the UI fast and treats a missing discovery Job as an idle/scheduled state rather than an exceptional condition.

  ## Open question

  If the sleep is still needed for the manual discovery trigger path, it may be better to handle that specifically instead of in the shared status-read method.

  I did not try to redesign that in this PR, because I wanted to keep the change small and focused on the UI latency issue.

  ## Testing
```
  go test ./ui -run TestGetRenovateJobs_ManyMissingDiscoveryJobsReturnsQuickly -count=1
  go test ./internal/renovate ./ui
  ```

### DIFF
--- a/src/internal/renovate/discoveryAgent.go
+++ b/src/internal/renovate/discoveryAgent.go
@@ -9,7 +9,6 @@ import (
 	"renovate-operator/internal/podLogs"
 	"renovate-operator/internal/types"
 	"sync"
-	"time"
 
 	"github.com/go-logr/logr"
 	"go.opentelemetry.io/otel"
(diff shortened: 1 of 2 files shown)


lukashankeln (approved):
Oh my god... Thank you for the fix 😄

PR #490, opened 2026-07-10
feat(operator): support username and TLS for host-based Valkey config

lauraseidler:
## What

  - Adds two optional environment variables to the host-based Valkey configuration:
    - `VALKEY_USERNAME` — connect as a specific ACL user; the built URL becomes `redis://user:pass@host:port/db`. A username without a password is emitted as `user@host` (for ACL users created with `nopass`).
    - `VALKEY_TLS` — when `true`, the built URL uses the `rediss://` scheme.
  - Both are ignored when `VALKEY_URL` is set, exactly matching the existing precedence contract for `VALKEY_HOST`/`VALKEY_PORT`/`VALKEY_PASSWORD`. 
  - Precedes the feature with a small refactor: `main.go` and `redisSecret.go` each built the same `ValkeyConfig` from env independently; this is now centralized in `kvstore.ConfigFromEnv`, with the lookup function injected so the `kvstore` package stays decoupled from the config singleton.
  - Fixes startup diagnostics: a failed Valkey connection previously exited with only `failed to initialize KV store` — the underlying error (TLS, auth, DNS, …) was discarded because `assert.Assert` doesn't print it. The error is now surfaced before exit.

  ## Why

  Managed Valkey offerings commonly deliver credentials as separate secret fields rather than a ready-made URL — for example the [Aiven operator](https://aiven.github.io/aiven-operator/)'s `ServiceUser` secret with `SERVICEUSER_HOST` / `SERVICEUSER_PORT` / `SERVICEUSER_USERNAME` / `SERVICEUSER_PASSWORD`. Without a username field, connecting as a non-default ACL user means templating a URL by hand outside the operator. With this change, the existing chart `extraEnv` mechanism covers it end to end — `docs/valkey.md` gains a complete example plus config-table entries for both new variables. No Helm template changes are needed.

  Verified against a real Aiven Valkey over GCP Private Service Connect (TLS with SNI routing, dedicated ACL user): operator startup, Valkey-backed session store, and the DB-offset cache URL in the forwarded job secret all work.

  **Known limitation (upstream, not this PR):** Renovate's own Redis client currently sends no TLS SNI, so executor jobs can't connect to SNI-routed `rediss://` endpoints (e.g. Aiven privatelink) even though the operator can — fix submitted as renovatebot/renovate#44469. Until that ships, `config.forwardCacheToJobs: false` is the workaround for such
  endpoints; sessions and log storage are unaffected.

### DIFF
--- a/charts/renovate-operator/templates/deployment.yaml
+++ b/charts/renovate-operator/templates/deployment.yaml
@@ -198,7 +198,58 @@ spec:
               value: {{ $githubRedirectUrl | quote }}
             {{- end }}
             {{- end }}
-            {{- if .Values.valkey.existingSecret }}
+            {{- $ekv := .Values.externalKeyValueStore }}
+            {{- if or $ekv.host $ekv.existingSecret.name }}
+            {{- if and $ekv.existingSecret.name $ekv.existingSecret.urlKey }}
+            - name: VALKEY_URL
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $ekv.existingSecret.name }}
+                  key: {{ $ekv.existingSecret.urlKey }}
+            {{- else }}
+            {{- if and $ekv.existingSecret.name $ekv.existingSecret.hostKey }}
+            - name: VALKEY_HOST
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $ekv.existingSecret.name }}
+                  key: {{ $ekv.existingSecret.hostKey }}
+            {{- else if $ekv.host }}
+            - name: VALKEY_HOST
+              value: {{ $ekv.host | quote }}
+            {{- end }}
+            {{- if and $ekv.existingSecret.name $ekv.existingSecret.portKey }}
+            - name: VALKEY_PORT
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $ekv.existingSecret.name }}
+                  key: {{ $ekv.existingSecret.portKey }}
+            {{- else if $ekv.port }}
+            - name: VALKEY_PORT
+              value: {{ $ekv.port | quote }}
+            {{- end }}
+            {{- if and $ekv.existingSecret.name $ekv.existingSecret.usernameKey }}
+            - name: VALKEY_USERNAME
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $ekv.existingSecret.name }}
+                  key: {{ $ekv.existingSecret.usernameKey }}
+            {{- else if $ekv.username }}
+            - name: VALKEY_USERNAME
+              value: {{ $ekv.username | quote }}
+            {{- end }}
+            {{- if and $ekv.existingSecret.name $ekv.existingSecret.passwordKey }}
+            - name: VALKEY_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $ekv.existingSecret.name }}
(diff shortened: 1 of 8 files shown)


lukashankeln:
I like the change 👍 

Should we add explicit Helm-Values to set the Valkey connection this way?

I just dont have a nice idea how to structure those values nicely without breaking the existing setup..
@wittdennis any ideas?

wittdennis:
Maybe we mark the current `existingSecret` & `existingSecretKey` deprecated and introduce a new top level key?:

```yaml
[...]
# -- redis protocol compatible key value store (e.g.: Redis / Valkey)
externalKeyValueStore:
  # -- whether the connection should be made with TLS encryption enabled
  useTls: false
  # -- host for the connection
  host: ~
  # -- port for the connection
  port: 6379
  # -- secret containing authentication secrets
  existingSecret:
    # -- name of the secret containing authentication information
    name: ~
    # -- key containing the full connection url. When set all other keys are ignored
    urlKey: ~ 
    # -- key containing the username for authentication
    usernameKey: ~
    # -- key containing the password for authentication
    passwordKey: ~

valkey:
   enabled: false
[...]
```

The new key would take precedence and we could remove the old ones in a future release. Would also give us a clear separation of concern between operator chart managed valkey and external systems.

wdyt?

lukashankeln:
> Maybe we mark the current `existingSecret` & `existingSecretKey` deprecated and introduce a new top level key?:
> 
> ```yaml
> [...]
> # -- redis protocol compatible key value store (e.g.: Redis / Valkey)
> externalKeyValueStore:
>   # -- whether the connection should be made with TLS encryption enabled
>   useTls: false
>   # -- host for the connection
>   host: ~
>   # -- port for the connection
>   port: 6379
>   # -- secret containing authentication secrets
>   existingSecret:
>     # -- name of the secret containing authentication information
>     name: ~
>     # -- key containing the full connection url. When set all other keys are ignored
>     urlKey: ~ 
>     # -- key containing the username for authentication
>     usernameKey: ~
>     # -- key containing the password for authentication
>     passwordKey: ~
> 
> valkey:
>    enabled: false
> [...]
> ```
> 
> The new key would take precedence and we could remove the old ones in a future release. Would also give us a clear separation of concern between operator chart managed valkey and external systems.
> 
> wdyt?

Sounds good! I agree, extracting this from the valkey key sounds like a good idea.

lauraseidler:
I like the design, and happy to add it to this PR, just one note - in some cases, host and port might _also_ be shipped via the secret (this is the case with the Aiven operator already, we are just not using those values as we use a different way of connection), and I could see that something like the username might _not_ be in some cases.

Should all of those keys support _either_ being set directly as `externalKeyValueStore.{param}`, or through `externalKeyValueStore.existingSecret.{param}Key`?

wittdennis:
IMO password and url must be set via an existing secret. That's a hard requirement.

The option to set every other key either as `externalKeyValueStore.{param}` or `externalKeyValueStore.existingSecret.{param}Key` (secret keys winning over clear values) would probably be most versatile, but I would be fine if we only implement one or the other.

lauraseidler:
@lukashankeln @wittdennis added as discussed - see if that looks good to you

wittdennis (approved):
Looks good 👍 

Thanks for the contribution!

PR #493, opened 2026-07-10
fix(helm): grafana dashboard label selector allow default overwrite

### LINKED ISSUE #485: Helm: Unable to strictly define matchLabels for GrafanaDashboard
**Describe the bug**
In the Helm chart values, defining a custom label for the grafanaDashboard (`metrics.dashboard.grafanaDashboard.instanceSelector.matchLabels`) is treated as an additional label to the default one `grafana.internal/instance: "grafana"`, as Helm merges the custom values with the default ones. In GrafanaDashboard resource, an instance should have all labels (and not at least one) specified in `matchLabels`. 

**Example**

If in my values I set : 
```yaml
    metrics:
      enabled: true
      serviceMonitor:
        enabled: true
      dashboard:
        enabled: true
        grafanaDashboard:
          allowCrossNamespaceImport: true
          instanceSelector:
            matchLabels:
              dashboards: grafana
```

the GrafanaDashboard resource spec ends up being like : 
```yaml
spec:                                                                                                                                                                                                                                                         
  allowCrossNamespaceImport: true
  configMapRef:
    key: renovate-operator.json
    name: renovate-operator-dashboard
  instanceSelector:
    matchLabels:
      dashboards: grafana
      grafana.internal/instance: grafana
```
which requires the Grafana instance to have both labels.

**Resolution ideas**

I like the idea of being able to pass the spec of the GrafanaDashboard in the values but it is also useful to have a default matchLabels section (for quick configuration). Maybe separate the matchLabels section of the entire spec to allow to handle the case separately in `charts/renovate-operator/templates/grafanadashboard.yaml` (like the hardcoded section for the `configMapRef`). If no matchLabels is set in values, set it to the default one otherwise strictly use the ones specified in the values.

lukashankeln:
fixes #485

### DIFF
--- a/charts/renovate-operator/templates/grafanadashboard.yaml
+++ b/charts/renovate-operator/templates/grafanadashboard.yaml
@@ -5,7 +5,17 @@ metadata:
   name: renovate-operator-dashboard
   namespace: {{ .Values.metrics.dashboard.namespace | default .Release.Namespace }}
 spec:
-  {{- toYaml .Values.metrics.dashboard.grafanaDashboard | nindent 2 }}
+  {{- toYaml (omit .Values.metrics.dashboard.grafanaDashboard "instanceSelector") | nindent 2 }}
+  instanceSelector:
+    {{- with omit (default (dict) .Values.metrics.dashboard.grafanaDashboard.instanceSelector) "matchLabels" }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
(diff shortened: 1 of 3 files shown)


lukashankeln:
@wittdennis is this breaking?! :/  What do you think...

wittdennis (approved):
I would say this is effectively non-breaking

PR #524, opened 2026-07-22
feat(helm): Allow overriding the webhook baseUrl

AaronLionsheep:
There are cases where the auto-derived value for the web hook route does not match the value that would be consumed by a public platform. Having an easy way to override the web hook's baseUrl solves this problem.


lukashankeln (approved):

PR #546, opened 2026-07-26
feat(operator): allow a per-job webhook base URL

MatthewStockdale:
### What

Adds an optional `spec.webhook.baseUrl` to the RenovateJob CRD. When set, it is used as the base for that job's synced webhook delivery URLs; when empty, the operator falls back to the existing `WEBHOOK_BASE_URL` environment variable.

### Why

One operator instance is explicitly designed to serve multiple Git platforms — one RenovateJob per platform, as described in `docs/webhooks/sync.md`. Webhook sync writes each hook's delivery URL from `WEBHOOK_BASE_URL`, which is process-wide: only the platform-specific *path* varies per job, the *host* is shared by all of them.

That breaks down when platforms need different hostnames to reach the same operator. A forge running inside the cluster can deliver to an internal-only address; a hosted platform such as github.com must reach a publicly resolvable one. With a single global base, operators have to either publish an otherwise internal receiver to the internet, or route in-cluster deliveries out through a public endpoint and back — both avoidable with a per-job override.

### How

- `RenovateWebhook` gains `BaseURL string` (`json:"baseUrl,omitempty"`, `+optional`).
- `webhookURLForJob` prefers `spec.webhook.baseUrl`, then `WEBHOOK_BASE_URL`; the platform path and the `?namespace=…&job=…` parameters are appended exactly as before. The "delivery URL is unknown" error is raised only when both are empty, and now names both.
- CRD regenerated with controller-gen v0.21.0.
- Unit tests for precedence, fallback, spec-only, and the updated error.
- Documented under "Webhook URL" in `docs/webhooks/sync.md`.

### Compatibility

Purely additive and backward compatible. `baseUrl` is optional; deployments that only set `WEBHOOK_BASE_URL` produce byte-identical delivery URLs. No Helm values change — the chart still derives `WEBHOOK_BASE_URL` from `webhook.route`/`webhook.ingress`, and the override lives on the RenovateJob.

### Testing

`go build ./...` and `go test ./...` pass. New tests in `src/internal/crdManager/renovateJobManager_test.go`.

Note: `golangci-lint` reports one pre-existing gofmt issue at `src/api/v1alpha1/renovatejob_types.go:223` (`ProjectStatus.Name` field alignment). It is present on `main` unmodified and unrelated to this change, so it was deliberately left out — happy to fold the one-line fix in here or send it separately if you prefer.


src/api/v1alpha1/renovatejob_types.go
```diff
@@ -115,7 +115,15 @@ type RenovateJobSecurityContext struct {
...
-	Enabled        bool                 `json:"enabled"`
+	Enabled bool `json:"enabled"`
+	// Externally reachable base URL of the operator's webhook server for this
+	// job, e.g. https://renovate.example.com. The platform-specific path is
+	// appended to it. Takes precedence over the operator-wide
+	// WEBHOOK_BASE_URL environment variable, which is used when this is empty.
+	// Set it when a platform needs a different hostname to reach the operator
+	// than the operator-wide default provides.
+	// +optional
+	BaseURL        string               `json:"baseUrl,omitempty"`
```
wittdennis, line 127:
```suggestion
    // +kubebuilder:validation:Pattern=`^https?://[^?#]+$`
	BaseURL        string               `json:"baseUrl,omitempty"`
```

Adding a rudimentary URL validation here would protect against basic misconfiguration

wittdennis (changes requested):
When webhook sync is active adding a baseUrl to the `RenovateJob` would cause a new webhook to be created as the match is the exact full url. If the old webhooks aren't manually removed this could potentially lead to a situation where renovate is fired twice, at least if both the old url and new url are reachable hosts. 

We should probably change the Url sync match for this. Maybe just match via path+query?

wdyt?

MatthewStockdale:
Agreed on both — pushed the validation pattern and the match change.

The way I'd frame it: this makes sync declarative. Right now identity and address are the same string, so changing the address makes the hook unrecognisable and we create a second one. Splitting them — identity is path + `job` + `namespace`, host is config — means the spec is desired state and we reconcile the hook to it instead of create-if-absent.

That needed one more change than the comparison itself. `ensureWebhook` short-circuited on `if hook.Active && hook.EventsUpToDate`, so a hook of ours on an old host would match, look healthy, and never get rewritten — worse than the duplicate, since there's no second hook to mask it and nothing in the logs. The URL now joins the drift check:

```go
if hook.URL == opts.WebhookURL && hook.Active && hook.EventsUpToDate {
    return nil
}
```

Match on identity, compare on full config. Identity is compared parsed rather than as a raw query string, since `q.Encode()` ordering is only guaranteed for hooks written by the operator. `removeWebhook` moves with it, otherwise we'd adopt hooks we then refuse to clean up.

The match fix stands on its own — a chart hostname or `baseUrlScheme` change breaks this today, without `baseUrl` existing at all — so it's a separate `fix(operator):` commit. Happy to split it into its own PR if you'd rather it landed independently, or got its own changelog entry.

New tests cover the migration path: a healthy hook under an old base URL is rewritten in place rather than duplicated, removal finds it too, another job's hook is left alone in both directions, and reversed parameter order still matches. The nine existing sync tests are unchanged and still pass.

Also updated the two places that documented the old identity model (`docs/webhooks/sync.md`, the architecture note in `CLAUDE.md`).

wittdennis (approved):
lgtm

Thanks for the contribution!

PR #547, opened 2026-07-27
feat(ui): collapse job cards by default and remember expansion

duderoot:
## What

Job cards on the dashboard now start collapsed instead of all being open.

When you have a few RenovateJobs each scanning a lot of repos, the dashboard
rendered every project of every job at once. With 5 jobs of 12 repos that's 60
rows and about 6 screens of scrolling — you couldn't even see which jobs exist
without scrolling past all of them. Collapsing by default keeps the overview
scannable and you open the one you actually care about.

Which cards you opened is now kept in `localStorage`, so it survives a reload
instead of snapping shut every time the page refreshes (which it does on its own
every 30s).

Also adds `Expand all` / `Collapse all`. They only act on the jobs you can
currently see, so if you've filtered by `Failed` or typed something in the search
box, expanding everything doesn't quietly open the jobs that are hidden.

## Tests

There was no frontend test coverage at all, so this adds a small Playwright setup
under `tests/ui` with four specs for the above.

The UI is fully client-side and gets everything from `/api/v1/*`, so the tests just
serve the real `index.html` and stub the API — no cluster and no Go build needed.
`staticFrontendServer.mjs` copies what `serveHTML` in `src/ui/ui.go` does with
`<base href>` and `window.__BASE_PATH__`, otherwise we'd be testing a page we never
actually ship.

They live outside `src/static` on purpose — the Dockerfile copies that whole
directory into the image, so anything in there gets shipped and served.

```sh
just test-ui                  # run them
just ui-dev                   # dashboard with mock data on :8098, no cluster
just test-ui-baseline HEAD    # run current specs against an older index.html
```

That last one is how I checked the specs aren't just passing by accident — against
`main` all four fail, the first one on `Expected: 0, Received: 60` rows.

The workflow to run this in CI is a separate PR, to keep this one to one thing.

## Checks

- `just test-ui` — 4 pass
- `just generate` — no CRD changes

README got the new recipes added to the *Running Tests* table. Nothing for `docs/`
since there's no new config, env var or Helm value.

## Notes

Only production file touched is `src/static/index.html` (+89 / −4), the rest is
tests and docs.

Also worth knowing: the mock API behind `just ui-dev` doesn't change state, so
hitting "run renovate" there returns 200 and nothing moves to Running. It's for
working on layout and client-side behaviour, not the job lifecycle.

All commits are signed off and the AI-assisted ones carry a `Co-Authored-By`.


lukashankeln:
I like the change but to me it feels like overkill to add playwright for 3 html files?

@beneiltis @wittdennis how do you feel about this?

wittdennis:
I'm with @lukashankeln. Change is good. Playwright is overkill for this. Although this is a potential point of improvement but one that I would like to address in a PR with the focus on that.

Don't know how I feel about the default collapsed. I think this will confuse users, especially new users that aren't familiar with the UI yet. I rather want the default to be expanded.

duderoot:
@lukashankeln and @wittdennis, thank you for the feedback.

The motivation for Playwright was the fact that the UI is at a stage where it feels like it's still at the beginning, and I can imagine that more features will be added over time.

Since I have "heard from a friend of mine" 😁that the UI is sometimes treated as a second-class citizen, I thought I would also add tests for it .

However, I respect your opinion, and I will update the PR and leave only the relevant changes. For the UI test/ Playwright , I will create a separate PR.

@wittdennis, regarding the default collapsed state, I can say that when you have 1,000+ projects, you're glad to have them collapsed. I understand your point, so the default expanded state works fine for me as well. I assume that the majority of users don't have that many projects.

wittdennis:
@duderoot with this many projects I can imagine that the UI will be pushed to its limits. I created issue #559 to explore ways to better support large deployments like yours. Feel free to add more info to the issue if you like

duderoot:
@wittdennis, thanks for the proposal in #559.

In this PR, I removed the entire Playwright infrastructure and left only the collapse/expand buttons, with the default state set to expanded.
Let me know if you'd like me to adjust anything else.

lukashankeln (approved):

PR #553, opened 2026-07-28
fix(rbac): narrow down secrets access

wittdennis:


charts/renovate-operator/templates/role/role.yaml
```diff
@@ -25,10 +25,12 @@ rules:
...
     verbs: ["get", "list"]
 
-  # Allow reading secrets for webhook token and github app integration
+  # Read platform/webhook tokens by name; create and update the github app token
+  # and valkey url secrets. No list/watch: Secrets bypass the informer cache, so a
+  # single grant here would otherwise expose every secret value in the namespace.
   - apiGroups: [""]
     resources: ["secrets"]
-    verbs: ["create", "get", "list", "watch", "update", "delete"]
+    verbs: ["get", "create", "update"]
```
lukashankeln, line 33: I think we might also need delete for example in case a Github App Connection gets replaced by a Token?

wittdennis: We don't actively delete any secret in the code base. The one place where we have a cleanup is when a Job is deleted by setting the owner reference of the github app token secret to that job. But that GC is actually done by the kube-controller-manager not by the operator

lukashankeln (approved):

beneiltis (approved):

PR #557, opened 2026-07-29
Security Hardening

wittdennis:
image

image

image

image

### DIFF
--- a/src/api/v1alpha1/renovatejob_types.go
+++ b/src/api/v1alpha1/renovatejob_types.go
@@ -55,16 +55,28 @@ type RenovateJobSpec struct {
 	SecurityContext *RenovateJobSecurityContext `json:"securityContext,omitempty"`
 	// Configuration for webhooks to trigger renovate runs
 	Webhook *RenovateWebhook `json:"webhook,omitempty"`
-	// Additional volumes to mount in the renovate pods
+	// Additional volumes to mount in the renovate pods.
+	// hostPath is rejected: it would give the pod the node's filesystem
+	// +kubebuilder:validation:MaxItems=64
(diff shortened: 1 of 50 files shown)


src/api/v1alpha1/renovatejob_types.go
```diff
@@ -55,16 +55,28 @@ type RenovateJobSpec struct {
 	SecurityContext *RenovateJobSecurityContext `json:"securityContext,omitempty"`
 	// Configuration for webhooks to trigger renovate runs
 	Webhook *RenovateWebhook `json:"webhook,omitempty"`
-	// Additional volumes to mount in the renovate pods
+	// Additional volumes to mount in the renovate pods.
+	// hostPath is rejected: it would give the pod the node's filesystem, and with it
```
lukashankeln: and with it?

src/internal/policy/policy.go
```diff
@@ -0,0 +1,392 @@
...
+	AllowRootUser bool
+	// AllowedImages are repository prefixes spec.image may resolve to.
+	// Empty denies every image.
+	AllowedImages []string
+}
+
+func FromConfig() Policy {
+	return Policy{
+		Disabled:                 config.GetValue("POLICY_ENABLED") == "false",
+		AllowedHosts:             parseList(config.GetValue("POLICY_ALLOWED_HOSTS")),
```
lukashankeln, line 94: From what I see I would have to explicitly enable my webhook url even though I have set it in the values as well?

The "risk factor" to me is the webhook url one can set in the job itself. The one configured in the chart should be allowed by default I would say?

lukashankeln (approved):
Well lets break stuff 😄

beneiltis:
break stuff *sing*

beneiltis (approved):

Issue #559, opened 2026-07-29
[Feature] UI: New RenovateJob overview page


wittdennis:
**Is your feature request related to a problem? Please describe.**
The operator UI gets quite cumbersome when you have a lot of projects. With 5 jobs of 12 repos that's 60
rows and about 6 screens of scrolling. 

**Describe the solution you'd like**
An overview page giving the essential and important information for each job without showing the full list of repositories. 

**Describe alternatives you've considered**
Setting the default visible state for all jobs to collapsed

wittdennis:
If you are running large deployments feel free to comment your pain points with the usability of the UI if you have any!

duderoot:
Make the header sticky so that filter information remains visible while scrolling. This issue becomes noticeable once the number of projects exceeds 10 to 20.

beisenherz:
The main difficulty for now is that the UI gets quite long. So having some setting that has it closed by default would be awesome. 
Also, maybe separate pages for different groups. 
Last, but not least, don't "forget" state (meaning not having the `x opened MRs` for only one run)

If you're having questions about this, feel free to ask. Love to help!

lukashankeln:
> Make the header sticky so that filter information remains visible while scrolling. This issue becomes noticeable once the number of projects exceeds 10 to 20.

That sounds like a good idea to me. But maybe we get it so that only the "row" with the filter badges is sticky and the "normal" header scrolls out of the way?

That would give us more space when scrolling while still leaving us with the information we need?

lukashankeln:
Another question in this category: Is the RenovateJob Card Header too big?

<img width="1255" height="209" alt="Image" src="https://github.com/user-attachments/assets/723a1dba-5f74-4019-8ed8-7c34f9619510" />

<img width="1268" height="98" alt="Image" src="https://github.com/user-attachments/assets/dacc453e-8f37-479c-a170-bf7fd4f83255" />

and mobile something like this?

<img width="294" height="69" alt="Image" src="https://github.com/user-attachments/assets/3839528f-369d-4fd8-918f-3d618592b3d8" />

With a lot of jobs this could also save a lot of space 🤔

ojsef39:
I'm running this multi-tenant: RenovateJobs generated per GitHub org via kustomize, each with its own renovate-$tenant group, plus renovate-admin on every job so I can see all of them.

[...]

PR #572, opened 2026-08-04
feat(ui): pin the dashboard filters and search to the top

duderoot:
On a dashboard several screens tall the stat filters, the search box and expand/collapse-all scrolled out of the viewport with everything else, so filtering something spotted far down the list meant scrolling back to the top and losing your place. It implements one part of #559 .

They now sit in a sticky toolbar directly under the brand strip, which scrolls away as before. 
grafik

The toolbar keeps the badges in one horizontally scrollable row and shortens the bulk-action labels on narrow screens, so it stays short enough not to eat the viewport it is pinned to. SiteHeader takes a hasBottomMargin prop so the bar can sit flush underneath it.


lukashankeln:
Looking good. I am not completely sure about the side-scrolling but I think that is a case of just trying it out and maybe iterating on it.

My question is, should we do the same with the logs-page? wdyt?

duderoot:
@lukashankeln the logs also have now a fixed header 😄 
Regarding the the side-scrolling  I don't understand what do you mean. Can you emphasize a bit?

lukashankeln (approved):
The Tools now scroll to the side, I am not yet sure about that one.

Image

Merging this as is though I will try this out a little bit in actual usage 👍 

Thank your for your contribution 😄

PR #582, opened 2026-08-06
Test/playwright UI suite

### LINKED ISSUE #581: Browser tests for the web UI: the frontend is the only part of the operator nothing verifies
**Problem:**
The operator's frontend has no automated coverage of any kind. `src/static/index.html`
is ~2,500 lines of untranspiled JSX that Babel compiles in the browser, plus
`pages/logs.html` next to it, and not one line of it is exercised by a test. The Go
side is well covered — ten `_test.go` files in `src/ui` alone — but those tests stop
at the HTTP boundary: they prove a route is registered, that auth redirects, that
`BASE_PATH` is normalised. None of them renders anything.

Everything a user actually touches lives on the other side of that boundary and is
therefore unverified:

- which job cards are expanded, and the `localStorage` bookkeeping that has to
  survive the 30s auto-refresh (otherwise a card the user closed springs back open)
- the stat filters, the project search, and the fact that "expand all" must act on
  the jobs those two leave visible rather than on every job
- each card's "Hide Projects" selection and what it persists
- the logs page's level filter, its persisted preference, and the `?levels=`
  override
- how the page consumes `/api/v1/logs`, which is an event stream and not a document

The practical consequences: 
- a frontend regression is found by a human clicking
around, or by a user; 
 - a fix to the UI cannot be shown to do what its commit message
claims; and shared markup cannot be refactored with any confidence, because there is
nothing to tell you the other page still renders the same thing. 
- On top of that,
looking at the UI at all currently means deploying to a cluster and waiting for a
real Renovate job to produce interesting state — so even manual verification is
expensive, and states like "a project that failed with three warnings and a truncated
PR list" are reached by luck rather than on purpose.

**Solution:**
A Playwright suite that drives the pages the operator actually ships, with no
cluster, no Go build, and no changes to the pages to make them testable.

- **Serve the real pages.** A small stand-in for the Go server that mirrors
  `serveHTML` and `registerUiRoutes`, including the `<base href>` and
  `window.__BASE_PATH__` it splices into `<head>`. A plain file server would silently
  test a page the operator never ships. `BASE_PATH=/renovate` runs the whole suite
  under a sub-path, which is otherwise only verified in Go unit tests.
- **Stub the API per spec** with `page.route`. The frontend gets everything from
  `/api/v1/*`, so route stubs are enough to drive it through any state — including
  ones a real cluster reaches rarely.
- **Fixtures that mirror the Go payloads** (`ui.RenovateJobInfo`,
  `crdManager.RenovateProjectStatus`, and the SSE frames `getRenovateJobLogs`
  writes), with a catalogue of every project state the dashboard renders differently,
  so a spec can name the states a filter should keep.
- **Live outside `src/static`.** The Dockerfile copies that whole tree into the image,
  so anything placed there would ship to production and be served.
- **Three recipes:** `just test-ui` to run the suite, `just ui-dev` to serve the
  frontend with mock data on its own port for manual work, and
  `just test-ui-baseline <rev>` to run today's specs against an older revision of the
  pages — which is how a new spec is shown to fail without the change it covers,
  rather than being assumed to.

**Describe alternatives you've considered**

- **Unit tests with jsdom / Testing Library.** There is nothing to import: each page
  is a single `<script type="text/babel">` block with no module boundaries and no
  exports. Getting them under a unit test runner means introducing a bundler and a
  build step into a frontend that is deliberately build-free, which is a much larger
  change than the problem justifies — and it would still not exercise the page the
  server serves.
- **Go integration tests driving the HTTP server.** They can assert what the API
  returns and that routing and auth behave, which is what the existing tests already
  do. They cannot see rendering, client-side state, or `localStorage`, which is where
  every one of the problems above lives.
- **Cypress.** Comparable capability. Playwright was chosen for first-class request
  stubbing (`page.route`), the trace viewer on failure, and one dependency with no
  additional runtime.
- **Visual regression / screenshot diffing.** Brittle against a CDN-built Tailwind and
  a browser-side Babel compile, and it answers "did the pixels change" rather than
  "does the behaviour hold". Worth adding later on top of this, not instead of it.
- **Carry on testing by hand.** The status quo. It does not scale past a couple of
  pages, it cannot be run by a reviewer, and it leaves every UI fix unprovable.

**Additional context**
- Cost: `npm ci` plus a Chromium download on first run; `node_modules`,
  `test-results` and reports are git-ignored. Nothing about the Go build, the image or
  the chart changes.

**Notice:**
Not covered by this request, and worth their own: wiring `just test-ui` into CI
(no workflow runs it today — `go-test.yaml` builds and tests Go only), and whether
to add a second viewport so the phone layout of the pinned toolbars is exercised.

duderoot:
TL;DR
According to #559 the UI is expecting to change and to grow by receiving new features. With the e2e test with playright we have a bigger confidence into changes 😜 

Closes #581.

## What

The harness that issue asks for, built as described: the real `index.html` and
`pages/logs.html` served through a stand-in for the Go server, `/api/v1/*` stubbed per
spec with `page.route`, fixtures mirroring the Go payloads, and everything under
`tests/ui` so nothing lands in the image.

**No production code is touched.** The diff is `tests/ui/**`, three recipes in the
`Justfile`, five ignore lines, and two lines of `CLAUDE.md` (the `tests/ui` tree and
`just test-ui` under *Verification*). Nothing about the Go build, the image, the chart
or the API changes, and neither page was modified to make it testable.

```sh
just test-ui                  # the suite, headless
just ui-dev                   # the frontend with mock data on :8098, no cluster
just test-ui-baseline <rev>   # today's specs against an older revision of the pages
```

## What it covers

16 specs, about 8 seconds, one Chromium project at 1280x900. They pass under
`BASE_PATH=/renovate` as well, which until now was only exercised by Go unit tests.

| spec file | what it pins |
|---|---|
| `jobCardExpansion.spec.mjs` | cards open per `JOB_CARDS_EXPANDED_BY_DEFAULT`; a collapse the user made survives the 30s auto-refresh; the bulk controls act only on the jobs the stat filter and the search box leave visible |
| `stickyToolbar.spec.mjs` | the dashboard's filters and search stay on screen while the brand strip scrolls away, stay usable from the bottom of the page, and sit below the overlay that dismisses a card's execution-options popover |
| `logsFiltering.spec.mjs` | the header counters follow the page's own level thresholds; the level filter narrows the list, persists, and yields to an explicit `?levels=`; search keeps the matching entries, highlights them, and opens the rows whose match is hidden below the message |
| `stickyLogsToolbar.spec.mjs` | the same three properties for the logs page's bar, including its level dropdown opening over the rows it floats above |

## How it works

- **The real pages, served the way the operator serves them.** `staticFrontendServer.mjs`
  mirrors `serveHTML` and `registerUiRoutes`, including the `<base href>` and
  `window.__BASE_PATH__` spliced into `<head>`. A plain file server would silently test
  a page that never ships.
- **The API is stubbed per spec** with `page.route`, so each test owns its data and can
  reach states a cluster produces rarely.
- **Fixtures mirror the Go payloads** — `ui.RenovateJobInfo`,
  `crdManager.RenovateProjectStatus`, and the SSE frames `getRenovateJobLogs` writes.
  `PROJECT_STATE_VARIANTS` is a catalogue of every project state the dashboard renders
  differently, so a spec can name the states a filter should keep instead of counting
  rows. When a field is added to either Go struct, it needs adding there too; the README
  says so, and the tests are only as honest as those payloads.
- **`/api/v1/logs` is an event stream**, so a spec serves the whole body in one
  `route.fulfill` and the browser's `EventSource` parses it frame by frame exactly as it
  would a live stream. `just ui-dev` writes the same frames on a timer instead, so
  clicking a log link shows the page streaming rather than a finished list.
- **It lives outside `src/static`** because the Dockerfile copies that tree into the
  image, so anything placed there would ship and be served.

## Does it actually catch anything?

That is the question a test-only PR has to answer, so `test-ui-baseline` exists to
answer it: it serves both pages from any revision while keeping today's specs.

Run against `9e894f1^` — the revision before the toolbar change of #572 landed:

```
5 failed
  3x sticky logs toolbar
  2x sticky dashboard toolbar
11 passed
  4x job card expansion
  6x logs filtering
  1x sticky dashboard toolbar  (the popover/stacking spec)
```

That split is the point. The five failures are the specs that pin what #572 changed —
they fail on the revision before it and pass on `main`. The eleven passes are specs
about behaviour that #572 did not touch, so they hold on both revisions and would have
caught a regression in it.

The one sticky spec that passes on both is deliberate and worth keeping: it guards a
regression that pinning the bar *could* introduce rather than proving a fix. With the
bar in the same stacking context as a card's execution-options popover, a click on a
stat badge would otherwise filter the list *and* leave the popover hanging over cards it
no longer belonged to.

## Decisions worth a second opinion

- **Log rows are located by their message and their level label**, not by a
  `data-testid`. A row has no role of its own, and I did not want to add test hooks to a
  production page for the sake of the suite. The cost: the level dropdown repeats those
  same words, so it has to be closed before counting rows — documented in the README and
  handled by the page object. If you would rather have `data-testid="log-row"` in the
  markup, that is a one-line change and I will make it.
- **One Chromium project at a fixed 1280x900.** A fixed viewport is what makes the
  above-the-fold assertions mean the same thing on every machine. The phone layout is
  not covered.
- **Two ways to produce a payload** — `page.route` stubs in specs, `mockOperatorApi.mjs`
  for `just ui-dev`. They share the fixture builders but not the wiring. The alternative
  was one mock server driving both, which would have made every spec depend on global
  state.

## Cost

`npm ci` plus a Chromium download on the first run of `just test-ui`; both are cached
afterwards. `node_modules`, `test-results` and the reports are git-ignored. `just build`,
`just test-unit`, `just test-helm` and `just generate` are unaffected — nothing in the
Go pipeline runs this.

## Not in this PR

Both are named as out of scope in #581 and I would rather they be judged on their own:

- **CI.** No workflow runs `just test-ui` today; `go-test.yaml` builds and tests Go only.
  Wiring it in, and deciding whether it gates merges, is the obvious next step.
- **A second viewport project**, so the phone layout of the pinned toolbars is exercised
  rather than spot-checked by hand.

## About the commit history

The branch is where the two toolbar fixes were developed, so `fix(ui): pin the dashboard
filters…` and `fix(ui): pin the logs filters…` appear in its history. Both are already
on `main` via #572, so they contribute nothing to this diff — `git diff origin/main...`
is tests only. There are also two merges from `main`. Happy to rebase or squash the
history into a single `test(ui):` commit if you prefer a clean list.

## Checklist

- [x] One logical change — the test harness and nothing else
- [x] Title follows Conventional Commits
- [x] Branch prefixed with the change type — `test/playwright-ui-suite` (`test` is a
      Conventional Commits type; `CONTRIBUTING.md` lists four prefixes and not this one)
- [x] Every commit carries a `Signed-off-by` (DCO)
- [x] AI-assisted commits carry a `Co-Authored-By` naming the model
- [x] No documentation impact — no new config, CRD, Helm value or user-facing setting;
      contributor docs are in `tests/ui/README.md` and two lines of `CLAUDE.md`


wittdennis:
Sorry for the late reply. Could you also add a ci job that runs the tests in PRs only when UI files changed

PR #602, opened 2026-08-14
feat(ui): enforce repository-level access

pat-s:
## What changed

- Filter dashboard projects through the authenticated user's Forgejo, Gitea, or GitHub repository access.
- Annotate projects with write permission and render read-only repositories separately.
- Require repository write access for individual and bulk Renovate triggers.
- Preserve OAuth access tokens in OIDC sessions so provider API permission checks can run.
- Integrate repository permissions with the v6 reader/admin authorization model.
- Treat missing Forgejo webhooks as successfully deleted and retain generated artifact compatibility after the upstream merge.

## Why

RenovateJob-level authorization alone can expose repositories that an authenticated user cannot access on the underlying Git provider.
The dashboard and mutation endpoints must also enforce the user's provider-level repository permissions.

## Impact

Authenticated Forgejo, Gitea, and GitHub users only see projects they can access.
Users need write permission on a repository to trigger Renovate for it, while readable repositories remain visible as read-only.
Deployments upgrading to v6 must configure `authorization` rules or set `authorization.enabled: false` to retain the previous authenticated-user behavior.

## Validation

- `just build`
- `just test-unit` (1,003 passed, 2 skipped)
- `just generate`
- `git diff --check`


pat-s:
Closing because this was opened against the operator source by mistake. The required fix belongs in the downstream Kubernetes deployment configuration.

PR #637, opened 2026-08-31
fix(chart): install every CRD from the hook, not just renovatejobs

pdecat:
## What

`crd.mode` defaults to `hook`, and the hook names one CRD file explicitly:

```yaml
data:
  renovatejob.yaml: |
{{ .Files.Get "crd/renovate-operator.mogenius.com_renovatejobs.yaml" | indent 4 }}
```

```yaml
            - -f
            - /crd/renovatejob.yaml
```

So the `RenovateProject` CRD added in #632 is shipped in the chart but never applied. `crd.mode: template` is unaffected, because `crds.yaml` already globs `crd/**`.

This is queued to ship: #625 (release 6.1.0) contains `extract project status into a separate crd per project`, `chart: adding permissions for renovateprojects` and `make crd project field immutable`. The permissions for the new kind landed in #634; this is the other half.

## What happens without it

The operator does not start. `controllers/renovateproject_controller.go` does `For(&api.RenovateProject{})` and is registered in `cmd/main.go`, so the manager tries to start an informer for a kind the API server does not serve, cache sync never completes and the pod crash-loops.

On a cluster where the CRD was never applied, the kind is simply absent:

```
$ kubectl apply -f renovateproject.yaml
error: resource mapping not found for name: "probe" namespace: "default":
no matches for kind "RenovateProject" in version "renovate-operator.mogenius.com/v1alpha1"
ensure CRDs are installed first
```

So a default install upgrading to 6.1.0 gets a crash-looping operator rather than a working one.

## The change

Glob `crd/**` into the ConfigMap and apply the mounted directory, so both install modes pick up whatever the chart ships and adding a CRD is not something to remember in a second place.

Three helm unittests cover it: both CRDs reach the ConfigMap, and the Job applies the directory rather than a named file. They fail against the template as it stands today (3 failed, 0 passed).

## Note on the ConfigMap keys

They change from `renovatejob.yaml` to the CRD filenames. The ConfigMap is a hook resource with `helm.sh/hook-delete-policy: before-hook-creation`, so it is recreated on every install and upgrade and nothing carries over.

I have not touched the committed CRDs. `just generate` currently rewrites ~125 lines of `renovatejobs.yaml`, which looks like drift from the Kubernetes 0.37 bump rather than anything to do with this, so it seemed better left out of a one-line template fix.

### DIFF
--- a/charts/renovate-operator/templates/crd-hook.yaml
+++ b/charts/renovate-operator/templates/crd-hook.yaml
@@ -51,8 +51,16 @@ metadata:
     helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
     helm.sh/hook-weight: "-5"
 data:
-  renovatejob.yaml: |
-{{ .Files.Get "crd/renovate-operator.mogenius.com_renovatejobs.yaml" | indent 4 }}
+  # Every CRD the chart ships, so adding one to crd/ is enough to have it
+  # installed. Naming files individually here means a new CRD is silently left
+  # uninstalled, and the operator only fails later, at the first API call that
+  # needs the missing kind. Mirrors the glob in crds.yaml, which the "template"
+  # install mode uses.
+{{- $files := .Files }}
+{{- range $path, $_ := .Files.Glob "crd/**" }}
+  {{ base $path }}: |
+{{ $files.Get $path | indent 4 }}
+{{- end }}
 ---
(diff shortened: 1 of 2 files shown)


wittdennis (approved):

lukashankeln:
Good catch, thanks for the contribution 😄

PR #652, opened 2026-09-02
feat(ui): info icon on job cards showing discovery config

simonkran:
## Summary

- Adds a small ⓘ icon to each RenovateJob card header in the dashboard
- On hover, a tooltip lists `discoveryFilters` and/or `discoverTopics` configured on the job
- Icon is only rendered when at least one of those fields is non-empty
- Exposes the two fields via the `/api/v1/renovatejobs` JSON response (`discoveryFilters`, `discoverTopics`)

## Test plan

- [ ] RenovateJob with `discoveryFilters` set → icon appears, tooltip shows filters
- [ ] RenovateJob with `discoverTopics` set → tooltip shows topics
- [ ] Job with both fields → both sections shown in tooltip
- [ ] Job with neither field → no icon rendered
- [ ] `just check` passes

### DIFF
--- a/src/static/index.html
+++ b/src/static/index.html
@@ -1754,6 +1754,14 @@
           key: "status",
           direction: "asc",
         });
+        const [infoTooltipPos, setInfoTooltipPos] = useState(null);
+        const infoIconRef = useRef(null);
+        const showInfoTooltip = () => {
+          if (!infoIconRef.current) return;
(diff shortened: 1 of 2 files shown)


beneiltis (approved):

wittdennis (changes requested):
Image

cut-off when job is collapsed

PR #654, opened 2026-09-02
fix(ui): info icon tooltip clipping and position

simonkran:
Follow-up to #652.

## Changes

- **Tooltip clipping fix**: switched from `position: absolute` (clipped by the card's `overflow: hidden`) to `position: fixed` calculated from `getBoundingClientRect()` on `mouseenter` — tooltip now renders correctly whether the job card is collapsed or expanded
- **Icon repositioned**: moved the ⓘ icon to sit immediately before the action buttons (Run Discovery / Trigger All), matching the location indicated in the design feedback

## Test plan

- [ ] Job with `discoveryFilters` or `discoverTopics` set: hover the icon while the card is **collapsed** — tooltip shows fully without clipping
- [ ] Same with card **expanded** — tooltip still shows correctly
- [ ] Icon appears between the Next Run column and the Run Discovery button
- [ ] `just check` passes

### DIFF
--- a/src/static/index.html
+++ b/src/static/index.html
@@ -1997,53 +1997,6 @@
                 </p>
               </div>
 
-              {/* Config info icon */}
-              {(job.discoveryFilters?.length > 0 || job.discoverTopics?.length > 0) && (
-                <div className="flex-shrink-0" onClick={(e) => e.stopPropagation()}>
-                  <svg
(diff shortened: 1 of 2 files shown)


simonkran:
I moved the information i icon next to 'Run Discovery'. Looks cleaner, IMO.

image

PR #661, opened 2026-09-03
fix(ui): keep filter summary visible while scrolling

benjaminknauer:
## What changed

Move the active filter summary into the sticky dashboard toolbar.

This keeps the filtered project and job counts, as well as the clear actions, visible while scrolling through long job lists.

## Testing

- `go test ./ui`
- Tested locally with multiple jobs and project statuses

### DIFF
--- a/src/static/index.html
+++ b/src/static/index.html
@@ -1154,6 +1154,48 @@
                   )}
                 </div>
               )}
+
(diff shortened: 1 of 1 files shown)


lukashankeln (approved):
Looking good 😄 

PR #666, opened 2026-09-07
fix: the redis/ valkey cache secret attaches to job lifecycle

### LINKED ISSUE #644: Forwarded redis cache isn't being rotated
**Describe the bug**
The renovate-operator creates a secret when you set it to forward the redis/valkey cache. This secret isn't being destroyed when no job needs it.

**To Reproduce**
1. Set operator to forward renovate secret
2. Run a renovate job
3. Rotate password
4. Rerun a renovate job
5. Renovate job fails with password error

**Expected behavior**
The operator creates 1 secret per job and deletes it afterwards or the password gets recreated every X minutes

 - Kubernetes Version [v1.35.6-eks-bca9cf6]
 - Version [6.0.1]

beisenherz:
Fixes #644 

Instead of one global redis url for all jobs, we can instead use the job lifecycle and give each job its own secret, which gets garbage collected as soon as the job is cleaned up.


lukashankeln (approved):
Looking good. Thanks for the contribution 😄

PR #678, opened 2026-09-11
Fix/626 trigger filtered projects

### LINKED ISSUE #626: Feature: Trigger filtered projects
**Is your feature request related to a problem? Please describe.**
Checking the UI, filtering projects for path/name etc `a/b/c`.
Then clicking "Trigger all" triggers ALL discovered repositories.

**Describe the solution you'd like**
I think it would be more UX-friendly if it would trigger only the filtered repositories.

**Describe alternatives you've considered**
-

best regards,
strowi

duderoot:
# fix(ui): trigger only the projects the dashboard's filters left visible

Closes #626.

## What

Filter the dashboard by name or path — `a/b/c` — and press **Trigger All**, and the
operator queues a run for *every* discovered repository. The filters only ever existed
in the browser: the request named no projects at all, so the handler updated every
project of the job.

The button now acts on the rows the card shows. Queued runs cannot be taken back, which
is what makes the old behaviour worth a fix rather than a docs note.

## The change

**Frontend** (`src/static/index.html`)

- `triggerAllRenovate(job, executionOptions, projects)` puts the visible project names
  in the request body.
- The scope covers all three filters the page offers: the stat badges and the search box
  (a new `projectListIsFiltered` prop from `App`) and the card's own **Hide Projects**
  menu, which a card already applies to its rows.
- The label states the scope — `Trigger All` becomes `Trigger 2` while a filter is
  active, with a matching `aria-label` and tooltip — so the narrowed action is visible
  *before* the click. The debug half of the split button follows the same scope, and the
  disabled state now keys off the filtered list rather than the raw one.

**Backend** (`src/ui/renovateController.go`)

- `POST /api/v1/renovate/all` accepts an optional `projects: []string` and restricts the
  batch update to those names.

```jsonc
// filtered — only these two are scheduled
{"renovateJob": "team-platform", "namespace": "renovate",
 "projects": ["acme/platform/api-docs", "acme/platform/api-gateway"]}

// unfiltered — unchanged meaning: every project the job has
{"renovateJob": "team-platform", "namespace": "renovate"}
```

## Compatibility

- **An absent or empty list keeps the old meaning.** Only an unfiltered dashboard omits
  the field, so "all" still includes repositories discovered *after* the page loaded,
  and any other API caller is unaffected.
- **No new access surface.** `triggerAll` is still required, and a named subset is
  strictly less than what that permission already grants. Names the job does not have
  are ignored rather than erroring — the list is a filter, not a create.
- **The running/scheduled skip is untouched**: a listed project that is already running
  stays untouched, as before.
- No CRD, Helm value, config option or user-facing setting changes.

## Tests

| test | what it pins |
|---|---|
| `tests/ui/specs/triggerAllScope.spec.mjs` (7) | unfiltered sends no list; search, stat filter and "Hide Projects" each scope the request to the visible rows; the label states the count; the debug button matches; a filter that empties a card disables it |
| `TestRunRenovateForAllProjects_ProjectScope` (5) | no list triggers everything; a list triggers only those; an empty list is treated as no list; a listed running project stays untouched; an unknown name triggers nothing else |

The specs have teeth: `just test-ui-baseline HEAD --grep "trigger all scope"` fails 7/7
against the pages as they are on `main`, and the Go table test fails 3/5 against the
unpatched handler.

**One fixture gap fixed along the way.** `buildRenovateJob` omitted `permissions`, `role`
and `accepted`, so every card the existing specs render was read-only with its action
buttons disabled — a spec could not have clicked one. It now mirrors `ui.RenovateJobInfo`
as `tests/ui/README.md` requires, and the project-name cell carries the `data-testid` the
fixture's own helper already expected.

## Verification

```
just check     # golangci-lint 0 issues, CRDs unchanged,
               # 1043 Go tests (2 skipped), helm-unittest 8 suites / 77 tests
just test-ui   # 23/23
```

## Commits

Two, so the fix can be cherry-picked without the tests:

- `fix(ui): trigger only the projects the filters left visible` — production code only
  (`index.html`, `renovateController.go`). Verified to build and pass `go test ./ui/` on
  its own.
- `chore(tests): cover the scope of the dashboard's trigger button` — specs, fixtures,
  `tests/ui/README.md`, Go handler test.

## Not in this PR

- **No `docs/` change.** The dashboard has no page of its own there, so there is nothing
  that describes the trigger button to correct. Happy to add one if you want the UI
  documented.

## Checklist

- [x] One logical change
- [x] Title follows Conventional Commits
- [x] Branch prefixed with the change type — `fix/626-trigger-filtered-projects`
- [x] Every commit carries a `Signed-off-by` (DCO)
- [x] AI-assisted commits carry a `Co-Authored-By` naming the model
- [x] `just check` passes locally
- [x] New behaviour covered by unit tests, both sides of the HTTP boundary

### DIFF
--- a/src/static/index.html
+++ b/src/static/index.html
@@ -684,6 +684,10 @@
             .filter((job) => job.projects.length > 0);
         }, [filteredJobs, searchTerm]);
 
+        // The stat badges and the search box narrow every card's project list, but a
+        // card cannot tell a filtered list from a short one — so App states it.
+        const projectListIsFiltered = Boolean(selectedStatFilter) || searchTerm.trim() !== "";
+
         const searchFilteredProjectCount = useMemo(() => {
           return searchFilteredJobs.reduce(
             (total, job) => total + (Array.isArray(job.projects) ? job.projects.length : 0),
@@ -969,7 +973,10 @@
           }
         };
 
-        const triggerAllRenovate = async (job, executionOptions = {}) => {
+        // `projects` is the list the caller wants run — the rows a card currently
+        // shows. Left null it is omitted from the request, which is what tells the
+        // operator to take every project of that job (issue #626).
+        const triggerAllRenovate = async (job, executionOptions = {}, projects = null) => {
           if (job.triggeringAll) return;
 
           setJobs((prev) =>
@@ -987,6 +994,7 @@
               body: JSON.stringify({
                 renovateJob: job.name,
                 namespace: job.namespace,
+                projects: projects && projects.length > 0 ? projects : undefined,
                 executionOptions: Object.keys(executionOptions).length > 0 ? executionOptions : undefined,
               }),
             });
@@ -994,8 +1002,10 @@
             if (response.ok) {
               addToast(
                 "success",
-                "All Triggered",
-                `All projects triggered for ${job.name}`
+                projects ? "Filtered Projects Triggered" : "All Triggered",
+                projects
+                  ? `${projects.length} ${projects.length === 1 ? "project" : "projects"} triggered for ${job.name}`
+                  : `All projects triggered for ${job.name}`
               );
             } else {
               const errorText = await response.text();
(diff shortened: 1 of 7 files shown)


lukashankeln (approved):
Looking good (test-errors are not related to your changes).

PR #686, opened 2026-09-17
fix(discovery): Propagate Renovate's exit code from the discovery job

PedDavid:
## What

`newDiscoveryJob` built its command ending in `&& cat repos.json || cat logs.json`. When Renovate fails, `cat` runs and succeeds, so the container exits `0` and the Job is marked `Complete`.

```sh
renovate --autodiscover --write-discovered-repos "$BASE_DIR/repos.json" >> "$BASE_DIR/logs.json" 2>&1 ||
  { rc=$?; cat "$BASE_DIR/logs.json"; exit $rc; }
cat "$BASE_DIR/repos.json"
```

`rc` is captured before `cat`, which would otherwise overwrite `$?`.

## Why

`ProcessDiscoveryJobResult` records a failure only when the Job itself failed, so the `failed` bucket of `renovate_operator_discovery_jobs_total` was unreachable — a discovery run that died during initialization looked identical to one that legitimately found nothing. The executor is unaffected: it runs `renovate` as the container command, so its exit code propagates, which is why `project_executions_total{status="failed"}` has always worked and its discovery counterpart never has.

Found with an expired GitLab token: Renovate logged `"Initialization error" errorMessage:"Authentication failure"`, the pod exited `0`, and the operator ran hourly discoveries that produced no projects for ten hours with nothing firing.

With this change the Job fails, so the metric increments and kube-prometheus-stack's `KubeJobFailed` fires without any new rule.

## Notes

- Success path unchanged; failure still prints `logs.json`, so `kubectl logs` remains the way to diagnose a run.
- `JOB_BACKOFF_LIMIT` defaults to `1800`, but `JOB_TIMEOUT_SECONDS` caps the whole Job at ~30 min, so a persistently failing discovery retries for at most that long. A lower backoff for discovery may be worth a separate look.
- Discovery logs still aren't parsed into `log_issues`, and there's no `fatal` bucket anywhere — separate follow-up.

## Tests

`TestNewDiscoveryJob_PropagatesRenovateExitCode` runs the real command string against a stubbed `renovate` on `PATH`, asserting exit code and stdout for the success and failure paths. This package has no existing `os/exec` precedent in tests — happy to swap for string assertions if you'd rather.

⚠️ Not built or tested locally (no Go toolchain on hand). Shell semantics were verified directly; the Go relies on CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lukashankeln (approved):
Looking good, thank you for the contribution 😄

PR #689, opened 2026-09-18
feat: adding keyboard navigation support for the ui

lukashankeln:


src/static/index.html
```diff
@@ -593,6 +607,12 @@
           persistStatFilter(null);
         }, []);
 
+        const [showShortcuts, setShowShortcuts] = useState(false);
+        const [focusedJobIndex, setFocusedJobIndex] = useState(null);
+        const [focusedProjectName, setFocusedProjectName] = useState(null);
+        const [focusedCardProjects, setFocusedCardProjects] = useState([]);
+        const jobCardRefsRef = useRef([]);
+
```
wittdennis, line 615: focusedJobIndex points at a position in searchFilteredJobs. When the search filter changes, the list gets shorter but the index stays where it is. So if you focus card 5 and then search for something that only returns 2 results, the index still says 5, which doesn't exist anymore. Pressing j or k next would resolve that, but until then Space, t, or ArrowDown just silently do nothing because they look up a card that isn't there. 

Reset focusedJobIndex to null when searchFilteredJobs changes (same for focusedRow/filteredLogs on the logs page).

wittdennis (changes requested):

beneiltis (approved):

PR #690, opened 2026-09-21
feat(discovery): select repositories by GitHub custom property (discoveryProperties)

aron-muon:
## What

`spec.discoveryProperties` on `RenovateJob`: select repositories by **GitHub repository custom properties** — the org-level metadata GitHub lets an organization require on every repository (e.g. an `owning-team` property).

```yaml
spec:
  discoveryProperties:
    - name: owning-team
      values: ["software-platform", "software-test"]
```

Right before each discovery run the operator lists the repositories the platform token can see and adds every one whose property matches to the discovery job's `RENOVATE_AUTODISCOVER_FILTER` (union with `discoveryFilters`). A repository joins or leaves a job by changing its property; the `RenovateJob` never carries a repository list. This is what lets one operator run one job per team across a whole organization, with `spec.access` handing each team admin of its own job.

## Why

Renovate's own autodiscover can filter by name pattern and topic, but not by custom property, and topics are free-form and per-repo while custom properties are org-governed (required, single-select, admin-controlled). Organizations that already classify repositories with properties get self-service onboarding into the right job for free.

## How

- `GitProviderClient.ListRepositoriesByProperty(ctx, name, values)`; the GitHub client pages `/installation/repositories` (App installation token) and falls back to `/user/repos` on 403/404 (PAT), reading the `custom_properties` map GitHub includes in repository objects. Single-select/text properties match on equality, multi-select when any selected value is listed; archived repositories are skipped. Bitbucket/Forgejo/Gitea/GitLab return `ErrCustomPropertiesUnsupported`.
- The discovery agent takes the provider client factory and resolves properties before building the discovery job (`withResolvedProjects`). **Fail closed**: if the properties are the job's only selection and resolve to nothing, discovery is refused instead of widened to the whole platform.
- CRD regenerated (`controller-gen`), manual `DeepCopyInto` extended, docs in `docs/configuration/autodiscovery.md`.

## Tests

- `githubProvider`: pagination across a full page, single/multi-select/null/absent properties, archived skip, PAT fallback, server error.
- `internal/renovate`: resolved projects join the filter env; resolution union + dedup, refusal when nothing resolves and no filters exist, unsupported platform error, no-op without properties.
- Existing mocks extended for the new interface method. `go vet` and `golangci-lint` clean; all packages green.

## Notes

GitHub only by nature of the feature; the field is documented as such. Happy to rename (`discoverByProperties`?) or to add per-property `exclude` semantics if you'd prefer.


aron-muon:
Hey @lukashankeln, what do you think about this change to allow filtering by custom repo property?

